Issue #99, opened 2014-12-12
AttributeError: '_UnpickleDispatch' object has no attribute '__name__'


lkrubner:
I was trying to pull some data from the database, using SqlAlchemy, and then turn them into JSON using jsonpickle, but as soon as I tried to call jsonpickle.encode(r) on the SqlAlchemy object, I got the following, very long, stack trace. 

AttributeError                            Traceback (most recent call last)
/Users/csr/projects/mn/repos/bronze/src/python2/privatepractice/export_from_legacy_to_bronze.py in <module>()
     21 
     22 if __name__ == '__main__':
---> 23     get_legacy_data_and_serialize()

/Users/csr/projects/mn/repos/bronze/src/python2/privatepractice/export_from_legacy_to_bronze.py in get_legacy_data_and_serialize()
     12     print("{} ".format(query.count()))
     13     for r in query.all():
---> 14         frozen = jsonpickle.encode(r)
     15         pprint(frozen)
     16         f = open('workfile' + str(i), 'w')

/Users/csr/projects/mn/repos/bronze/nixd/opt/python2.7-privatepractice/lib/python2.7/site-packages/jsonpickle/__init__.pyc in encode(value, unpicklable, make_refs, keys, max_depth, backend, warn, max_iter)
    129                           keys=keys,
    130                           max_depth=max_depth,
--> 131                           warn=warn)
    132 
    133 

/Users/csr/projects/mn/repos/bronze/nixd/opt/python2.7-privatepractice/lib/python2.7/site-packages/jsonpickle/pickler.pyc in encode(value, unpicklable, make_refs, keys, max_depth, reset, backend, warn, context, max_iter)
     39                           warn=warn,
     40                           max_iter=max_iter)
---> 41     return backend.encode(context.flatten(value, reset=reset))
     42 
     43 

/Users/csr/projects/mn/repos/bronze/nixd/opt/python2.7-privatepractice/lib/python2.7/site-packages/jsonpickle/pickler.pyc in flatten(self, obj, reset)
    139         if reset:
    140             self.reset()
--> 141         return self._flatten(obj)
    142 
    143     def _flatten(self, obj):

/Users/csr/projects/mn/repos/bronze/nixd/opt/python2.7-privatepractice/lib/python2.7/site-packages/jsonpickle/pickler.pyc in _flatten(self, obj)
    143     def _flatten(self, obj):
    144         self._push()
--> 145         return self._pop(self._flatten_obj(obj))
    146 
[...]

PR #100, opened 2014-12-12
I ran into some problems which I detailed in issue #99

lkrubner:
I ran into some problems which I detailed in issue 99, https://github.com/jsonpickle/jsonpickle/issues/99. I was pulling models out of SqlAlchemy and some of the sub-classes were not iterable, so I went through and added if/else blocks where I was facing problems. This now works for me, without any obvious errors.

### DIFF
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -348,8 +348,9 @@ class Pickler(object):
             if insufficiency:
                 rv_as_list+=[None]*insufficiency
 
-            if rv_as_list[0].__name__ == '__newobj__':
-                rv_as_list[0] = tags.NEWOBJ
+            if hasattr(rv_as_list[0], __name__): 
+                if rv_as_list[0].__name__ == '__newobj__':
+                    rv_as_list[0] = tags.NEWOBJ
 
             data[tags.REDUCE] = list(map(self._flatten, rv_as_list))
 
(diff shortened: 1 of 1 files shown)


abhijeetkalyan:
Hi, just wondering what the status of this PR was. Having the same issue as @lkrubner

lkrubner:
I am sorry I only noticed your comments now. I don't often work with Python, so I only catch up on Python projects maybe once a year. I am sorry to say I no longer have the original problem in front of me. But I recall this line gave me several errors: 

       if rv_as_list[0].__name__ == '__newobj__':

And so the only way I could get this to work was wrapping it with: 

      if hasattr(rv_as_list[0], __name__):

davvid:
I'd like to see this expressed in a unit test form. Can you please share a snippet that we can keep around in the test suite for validating this use case and potential fix? The best part is that if we have a sufficiently elaborate test case then you can be sure that we won't ever break your use case in the future.

The `str(...)` casts also caught my attention -- it made me wonder about `float` and `int` fields, so I'm curious to get a good test case for this.  Keeping sqlalchemy-related tests in the test suite is fine with me. When testing we can use a `sqlite:///:memory:` url to allow the tests to not rely on the filesystem.

davvid:
I added sqlalchemy tests to our test suite so I went ahead and closed this issue.  The tests in `tests/sqlalchemy_test.py` are basically what we can currently support.

The tests that were added may be overly simplistic. If you're able to demonstrate a breakage please try to do it by adding new tests to `tests/sqlalchemy_test.py` so that we can have a common baseline to work against.

Thanks again for the report. Let us know if there's anything more we should add to the existing tests that would help your use case.

PR #105, opened 2014-12-29
Basic support for pickling types with metaclasses

aldanor:
This now works out of the box on both Python 2 and Python 3:
```python
>>> import enum
>>> class Foo(enum.IntEnum):
...     A = 1
...     B = 2
... 
>>> import jsonpickle
>>> jsonpickle.encode(Foo.A)
'{"py/reduce": [{"py/type": "__main__.Foo"}, {"py/tuple": [1]}, null, null, null], "py/object": "__main__.Foo", "py/newargs": {"py/tuple": [1]}}'
>>> jsonpickle.decode(jsonpickle.encode(Foo.A))
<Foo.A: 1>
```
There were two problems about this: `has_get*` flags in `_flatten_obj_instance` were not set correctly and `util.is_type` was returning false for types with metaclasses.

There may be some unobvious edge cases with twisted usage of metaclasses, but it at the very least seems to handle `enum` which is not the simplest example of all (https://hg.python.org/cpython/file/3.4/Lib/enum.py).

### DIFF
--- a/jsonpickle/util.py
+++ b/jsonpickle/util.py
@@ -15,6 +15,7 @@ import io
 import operator
 import time
 import types
+import inspect
 
 from jsonpickle import tags
 from jsonpickle.compat import set
@@ -50,12 +51,48 @@ def is_type(obj):
         return isinstance(obj, (type, types.ClassType))
 
 
-def has_method(obj, name, *args, **kwargs):
-    try:
-        getattr(obj, name)(*args, **kwargs)
-        return True
-    except:
+def has_method(obj, name):
+    # false if attribute doesn't exist
+    if not hasattr(obj, name):
         return False
+    func = getattr(obj, name)
+
+    # builtin descriptors like __getnewargs__
+    if isinstance(func, types.BuiltinMethodType):
+        return True
+
+    # note that FunctionType has a different meaning in py2/py3
+    if not isinstance(func, (types.MethodType, types.FunctionType)):
+        return False
+
+    # need to go through __dict__'s since in py3 methods are essentially descriptors
+    base_type = obj if is_type(obj) else obj.__class__  # __class__ for old-style classes
+    original = None
+    for subtype in inspect.getmro(base_type):  # there is no .mro() for old-style classes
+        original = vars(subtype).get(name)
+        if original is not None:
+            break
+
+    # name not found in the mro
+    if original is None:
+        return False
+
+    # static methods are always fine
+    if isinstance(original, staticmethod):
+        return True
+
+    # at this point, the method has to be an instancemthod or a classmethod
+    self_attr = '__self__' if PY3 else 'im_self'
+    if not hasattr(func, self_attr):
+        return False
+    bound_to = getattr(func, self_attr)
+
+    # class methods
+    if isinstance(original, classmethod):
+        return issubclass(base_type, bound_to)
+
+    # bound methods
+    return isinstance(obj, type(bound_to))
 
 
 def is_object(obj):
--- a/jsonpickle/util.py
+++ b/jsonpickle/util.py
@@ -43,10 +43,19 @@ def is_type(obj):
     >>> is_type(Klass)
     True
     """
+    # use "isinstance" and not "is" to allow for metaclasses
     if PY3:
-        return type(obj) is type
+        return isinstance(obj, type)
     else:
-        return type(obj) is type or type(obj) is types.ClassType
+        return isinstance(obj, (type, types.ClassType))
+
+
+def has_method(obj, name, *args, **kwargs):
+    try:
+        getattr(obj, name)(*args, **kwargs)
+        return True
+    except:
+        return False
 
 
 def is_object(obj):
@@ -62,8 +71,7 @@ def is_object(obj):
     False
     """
     return (isinstance(obj, object) and
-            type(obj) is not type and
-            type(obj) is not types.FunctionType)
+            not isinstance(obj, (type, types.FunctionType)))
 
 
 def is_primitive(obj):
@@ -82,6 +90,7 @@ def is_primitive(obj):
         return True
     return False
 
+
 def is_dictionary(obj):
     """Helper method for testing if the object is a dictionary.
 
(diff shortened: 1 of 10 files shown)


jsonpickle/util.py
```diff
@@ -43,10 +43,19 @@ def is_type(obj):
...
-        return type(obj) is type or type(obj) is types.ClassType
+        return isinstance(obj, (type, types.ClassType))
+
+
+def has_method(obj, name, *args, **kwargs):
+    try:
+        getattr(obj, name)(*args, **kwargs)
+        return True
+    except:
+        return False
```
davvid: It looks like this `has_method()` predicate also has the side-effect of actually calling the method. That can have strange side-effects in user code and it's a bit wasteful.

How about reworking this to check whether the attribute is callable. You'll need to adjust the tests, but maybe something like:

    def has_callable(obj, name):
        return callable(getattr(obj, name))

...and then use `has_callable()` instead of `has_method()`.  When the attribute doesn't exist we'll get `None` from `getattr()` and `callable()` will return `False`.

davvid: Hmm I see your earlier message mentioning that it only fails at runtime because `__getnewargs__` needs an `int` argument so I see the reasoning for actually wanting to call the method to see if it actually works. That's a bummer.

Is there any way to encode/know exactly what that `int` arg should be so that we don't have to guess?

Also, it would be good for the test cases to include an actual Enum test case since that's the problem we're trying to fix. Verifying that the actual enum value makes it through the round-trip is what we should verify.

aldanor: Your point is absolutely valid, calling a method to check if it's callable is a very cheap way of doing it :) Basically, what we we want to check is that if the attribute points to a bound method, we want the object to be an instance and not a type (which is exactly why it was failing with enum).

Now that I think about it, you can probably get away with something like this:
```python
if getattr(obj, attr).im_self is not None
```

aldanor: Another thing about the enum case is that `Foo.A` is both an instance of `Foo`, a type, and an `int`, all at the same time... I hope the above fix will suffice, though.

aldanor: Re: the enum test case, I'd be happy to add one, if you don't mind an extra test dependency for py2 (enum34 on pypi). Could add a tox recipe as well if you want.

davvid: Thanks @aldanor , that makes sense.  I guess if we can get away with not calling the method then that's ideal, but if it doesn't work out then I'm okay being pragmatic.

RE: the dependency thing -- I'd like to avoid the dependency if possible. If you `git grep "if PY3" -- tests` you'll see a few cases where we skip tests based on the python version, e.g.:

        if PY3:
            self.skipTest('No classic classes in PY3')

Maybe the tests could only run if `enum` is importable -- that way it's a soft dependency and won't force users to install the `enum` module on py2.

davvid: This enum object is pretty interesting :-) poking at it a bit, I did notice that `Foo.A.__getnewargs__.__self__` exists, so maybe the `__self__` attribute can give us additional clues. I'm mentioning it just in case it can help us out in this situation. Excellent sleuthing.

davvid: Oh yeah, sorry for missing your note -- yes, a tox recipe would be very cool. I've not used tox much myself (yet), but it sounds like I really should. I personally have a shell script that jumps into different virtualenvs and runs the JP unit tests using different pythons, but tox seems like a much nicer and official way to do that. :+1:

aldanor: @davvid Alright then, I'll do it :)

Another minor thing I've noticed -- have you considered replacing the boilerplate in `jsonpickle.compat` with `six` so you don't have to reinvent the wheel? (as recommended in the official python docs --https://docs.python.org/3/howto/pyporting.html#drop-support-for-python-2-5-and-older-at-least). It's a slim requirement and the chances are most users already have it installed due to it being a very common dependency.

Otherwise, we'll have to add even more unneeded copy/paste style boilerplate to `compat` in order to have proper metaclass unit tests in both python 2/3... It also provides things like `six.class_types`, `six.get_unbound_method` which are done differently in 2/3 and could be handy here.

davvid:
```
$ python
Python 2.7.5 (default, Mar  9 2014, 22:15:05)
[GCC 4.2.1 Compatible Apple LLVM 5.0 (clang-500.0.68)] on darwin
>>> import six
ImportError: No module named six
```

I know this sounds silly but please avoid the dependency if possible. I know it's a bit of copy/paste boilerplate, but that boilerplate is not changing anytime soon since the 2/3 overlap is a relatively known beast.  compat is there to avoid the dependency and is only 42 lines right now.

aldanor:
@davvid take #2, have a look? :)

`has_method` is quite hardcore now due to required support for both old style classes and Python 3 (where methods have becomes descriptors and pass the `types.FunctionType` check...). The tests seem to pass though on py26/27/33/34.

// I'm not quite sure I've done rebasing right -- I've rebase pulled upstream master into my branch

aldanor:
I forgot to mention, there is now a `tox` recipe which replaces the following files that can be now removed if you want:
`requirements-2.txt`, `requirements-3.txt`, `requirements-test.txt`, `requirements.txt`, `Rakefile`, `tests/tests-req.txt`.

To run tests locally, just run `tox`, that should do it. The virtual environments are automatically cached and/or cleaned up. To run tests for a specified environment(s), you can do something like `tox -e py26 -e py27`. You can also pass args to nosetests like so `tox -- --tests=util_test -v -s`.

I've updated the travis build config to use tox as well which seems to work.

davvid:
Awesome, thank you.  I pulled your branch down, rebased locally (to get rid of the duplicate commits) and merged this in (so don't be alarmed if this says "Closed" ~ it's actually merged).

Nicely done.  I think we're just about ready for the v0.8.1 release.  What do you think @marcintustin ?

marcintustin:
I think so.

-- 
Marcin Tustin
Tel: +44 (0) 7773 787 105 (UK)
       +1  917 553 3974 (US)

aldanor:
@marcintustin @davvid Btw: I can add some enum tests later tonight if you want.

And the `ext.numpy` stuff would go into the next release.

davvid:
I'll always take more tests ;-) especially before a release

Issue #108, opened 2015-01-09
Encoding byte strings in Python 3 causes infinite recursion


aldanor:
```python
>>> jsonpickle.encode(b'foo')
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-7-5e089e18bf41> in <module>()
----> 1 jsonpickle.encode(b'foo')

/Users/ivansmirnov/dev/jsonpickle/jsonpickle/__init__.py in encode(value, unpicklable, make_refs, keys, max_depth, backend, warn, max_iter)
    129                           keys=keys,
    130                           max_depth=max_depth,
--> 131                           warn=warn)
    132 
    133 

/Users/ivansmirnov/dev/jsonpickle/jsonpickle/pickler.py in encode(value, unpicklable, make_refs, keys, max_depth, reset, backend, warn, context, max_iter)
     39                           warn=warn,
     40                           max_iter=max_iter)
---> 41     return backend.encode(context.flatten(value, reset=reset))
     42 
     43 

/Users/ivansmirnov/dev/jsonpickle/jsonpickle/pickler.py in flatten(self, obj, reset)
    139         if reset:
    140             self.reset()
--> 141         return self._flatten(obj)
    142 
    143     def _flatten(self, obj):

/Users/ivansmirnov/dev/jsonpickle/jsonpickle/pickler.py in _flatten(self, obj)
    143     def _flatten(self, obj):
    144         self._push()
--> 145         return self._pop(self._flatten_obj(obj))
    146 
    147     def _flatten_obj(self, obj):

/Users/ivansmirnov/dev/jsonpickle/jsonpickle/pickler.py in _flatten_obj(self, obj)
    159             return None
    160 
--> 161         return flatten_func(obj)
    162 
    163     def _list_recurse(self, obj):

/Users/ivansmirnov/dev/jsonpickle/jsonpickle/pickler.py in _ref_obj_instance(self, obj)
    214             # We've never seen this object so return its
    215             # json representation.
--> 216             return self._flatten_obj_instance(obj)
    217         # We've seen this object before so place an object
    218         # reference tag in the data. This avoids infinite recursion

...

/Users/ivansmirnov/dev/jsonpickle/jsonpickle/pickler.py in _flatten(self, obj)
    143     def _flatten(self, obj):
    144         self._push()
--> 145         return self._pop(self._flatten_obj(obj))
    146 
    147     def _flatten_obj(self, obj):

[...]

Issue #101, opened 2014-12-16
jsonpickle and numpy float enter recursive loop and raise RuntimeError


ianozsvald:
Using jsonpickle on a pure-Python float works fine, doing the same on a numpy.float_ causes a RuntimeError to be raised due to a recursion limit. The same behaviour can be seen if using numpy.int_ and numpy.bool_.

This sample code reproduces the bug:
```
"""jsonpickle fails with recursion error on numpy floats"""
# Using Python 3.4 via Anaconda
import jsonpickle  # version == 0.8.0
import numpy as np  # __version__ == 1.9.1

# make a numpy float object
f = np.float_(42)

# encoding a pure-Python float works correctly
f_float_encoded = jsonpickle.encode(float(f))

# fails with:
# RuntimeError: maximum recursion depth exceeded while calling a Python object
f_numpy_encodoed = jsonpickle.encode(f)  # raises RuntimeError
```

Tail end of the stack trace:
```
/home/ian/anaconda/envs/elevate_direct/lib/python3.4/site-packages/jsonpickle/pickler.py in _ref_obj_instance(self, obj)
    211             # We've never seen this object so return its
    212             # json representation.
--> 213             return self._flatten_obj_instance(obj)
    214         # We've seen this object before so place an object
    215         # reference tag in the data. This avoids infinite recursion

/home/ian/anaconda/envs/elevate_direct/lib/python3.4/site-packages/jsonpickle/pickler.py in _flatten_obj_instance(self, obj)
    287 
    288             if has_getnewargs:
--> 289                 data[tags.NEWARGS] = self._flatten(obj.__getnewargs__())
    290 
    291             if has_getinitargs:

/home/ian/anaconda/envs/elevate_direct/lib/python3.4/site-packages/jsonpickle/pickler.py in _flatten(self, obj)
    143     def _flatten(self, obj):
    144         self._push()
--> 145         return self._pop(self._flatten_obj(obj))
    146 
    147     def _flatten_obj(self, obj):

/home/ian/anaconda/envs/elevate_direct/lib/python3.4/site-packages/jsonpickle/pickler.py in _flatten_obj(self, obj)
    159             return None
    160 
--> 161         return flatten_func(obj)
    162 
    163     def _list_recurse(self, obj):

/home/ian/anaconda/envs/elevate_direct/lib/python3.4/site-packages/jsonpickle/pickler.py in <lambda>(obj)
    182             if not self.unpicklable:
    183                 return list_recurse
[...]

Issue #13, opened 2010-12-14
jsonpickle fails to preserve Unicode strings


gnosek:
Any chance of decode("foo") returning a Unicode string? E.g. controlled by a parameter, not to break compatibility.

```
Python 2.6.5 (r265:79063, Apr 16 2010, 13:09:56) 
[GCC 4.4.3] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from jsonpickle import encode, decode
>>> type(decode(encode(u'foo'))) == type(u'foo')
False
```

PR #110, opened 2015-01-09
Proper bytestring support + numpy extension

aldanor:
This adds support for bytestring encoding/decoding as discussed in #108, plus numpy support with some tests.

I don't think the unicode/bytes change is bound to break much since
- unicode strings in both Python 2/3 are encoded and decoded same as before, via json literals; in addition, unicode in always means unicode out
- encoding of bytestrings that are not decodable into utf-8 in Python 2 were failing anyway (downstream in the backend code, for most or all backends), so there's nothing to break there
- the most significant change is proper encoding of Python 3 bytestrings -- but then again, (a) Python 3 makes a hard distinction between bytes/unicode so the chances are one would know what he's doing if using them, (b) trying to encode a bytestring in py3 as of the current master would raise a recursion limit exception (#101, #108) so noone would be possibly doing that

The reason for having unicode as default and bytes as a special case (and not the other way around as suggested in #13) is that (a) the default text type in Python 3 is unicode, and bytes is a essentially a binary type, (b) bytestrings natively pickled in Python 2 are decoded as unicode in Python 3, by design.

The only thing that's lacking is being able to force encoding ascii bytestrings as true bytestrings in Python 2, but I don't see an easy way to do this without overcomplicating the whole thing.

There may still be a few things to add or polish but I wanted to open a PR early for @davvid and @marcintustin to take a look. For one thing, there need to be additional tests for the whole bytes/unicode logic (edit: *done*). Maybe some additional numpy tests as well to verify the encoding format explicitly and not only the roundtrip.

On a sidenote, `jsonpickleJS` would probably need to be updated as well to be aware of `py/bytes` and `quopri` encoding.


aldanor:
Related: #101, #108.

davvid:
The code is good but after merging the tests fail because of doctests:

```
$ ./tests/runtests.py
File "jsonpickle/pickler.py", line 118, in jsonpickle.pickler.Pickler.flatten
Failed example:
    p.flatten('hello world')
Expected:
    'hello world'
Got:
    u'hello world'
File "jsonpickle/pickler.py", line 137, in jsonpickle.pickler.Pickler.flatten
Failed example:
    p.flatten({'key': 'value'})
Expected:
    {'key': 'value'}
Got:
    {'key': u'value'}
```

I remember these docstrings. I've had to fix them before when porting to python3, and I'm pretty sure fixing them for py2 breaks py32, py33 or something.  IMO these two docstrings have been too high maintenance and are better off removed.  Want to do the honors?

aldanor:
Tests work in tox for me locally and they work on travis, strange. What interpreter version are you running them from, wonder what exactly is different

Re: doctests,

aldanor:
Oh, ok, I see -- fails for me as well.

Issue #117, opened 2015-03-01
Object mismatch with pymongo FixedOffset


jaraco:
I've discovered that the round-trip serialization/deserialization of a simple FixedOffset object produces a corrupted object. The output of [this test script](https://gist.github.com/jaraco/d979a558bc0bf2194c23/079efd37a641159c0c14a1ab93f7efce30bf3eda) produces:

```
============================= test session starts =============================
platform win32 -- Python 3.4.2 -- py-1.4.26 -- pytest-2.6.4
collected 1 items

test_fo.py F

================================== FAILURES ===================================
________________________ test_FixedOffsetSerializable _________________________

    def test_FixedOffsetSerializable():
    	import jsonpickle
    	import bson.tz_util
    	fo = bson.tz_util.FixedOffset(-60*5, 'EST')
    	serialized = jsonpickle.dumps(fo)
    	pprint.pprint(json.loads(serialized))
    	restored = jsonpickle.loads(serialized)
    	print(restored._FixedOffset__offset)
>   	assert restored == fo
E    assert <bson.tz_util.FixedOffset object at 0x0000000005F494A8> == <bson.tz_util.FixedOffset object at 0x0000000005EEFA90>

test_fo.py:16: AssertionError
---------------------------- Captured stdout call -----------------------------
{'_FixedOffset__name': 'EST',
 '_FixedOffset__offset': {'py/id': 1},
 'py/initargs': {'py/tuple': [{'__reduce__': [{'py/type': 'datetime.timedelta'},
                                              {'py/tuple': [-1, 68400, 0]}],
                               'py/object': 'datetime.timedelta'},
                              'EST']},
 'py/object': 'bson.tz_util.FixedOffset'}
<bson.tz_util.FixedOffset object at 0x0000000005F494A8>
========================== 1 failed in 0.52 seconds ===========================
```

As you can see, the object produced isn't equal to the one serialized. Furthermore, it appears as if the `.__offset` property of the object has been set to itself instead of the timedelta object.

marcintustin:
Does it roundtrip correctly with pickle? I wonder if there are two problems here - one with the pickling machinery they provide, and one with ours.

jaraco:
The roundtrip does work correctly with stdlib pickle. Here's [rev 3 of the test script](https://gist.github.com/jaraco/d979a558bc0bf2194c23/4391e5ea2ae8ac8ace31904e4799b3dc49fada10) where the first test still fails, but the other two succeed.

[...]

PR #119, opened 2015-03-01
Some minor cleanup

jaraco:
In my debugging of #117, I encountered some patterns in the code that I thought could be improved. Please consider pulling in these changes, which should have no behavioral changes, but are clearer to read.


marcintustin:
I'm happy with this.

Issue #121, opened 2015-03-15
Handling a deque with maxlen


sirpercival:
Hi all,

jsonpickle doesn't play very nicely with deque if you include a maxlen parameter. I tried to use the fix shown [here on stackoverflow](http://stackoverflow.com/questions/19494734/jsonpickle-ignores-maxlen-of-collection-deque):
```python
class DequeHandler(jsonpickle.handlers.BaseHandler):
    def flatten(self, obj, data):
        return list(obj),obj.maxlen
    def restore(self,obj):
        return deque(obj[0], maxlen=obj[1])
jsonpickle.handlers.register(deque, DequeHandler)
```
 but now the deque is being decoded into a list: [[data], maxlen]. Any suggestions for how to improve the handler? And maybe it might be nice to include something like this in handlers.py?

Thanks!

sirpercival:
Actually, SimpleReduceHandler does perfectly well with deques. I'll do a pull request.

davvid:
Thanks for the fix, merged in da2dd46d3540a6c1c1da83c2ab8b300882d3f769

PR #122, opened 2015-03-15
Add deque maxlen support

sirpercival:
Using SimpleReduceHandler preserves the maxlen parameter of deques on decode, as per #121.


davvid:
I'd hate to break your use cases in the future.  Can you please add a test case (you can run the test suite by running `./tests/runtests.py`) to `tests/object_test.py` that exercises your use case? That way we'll be protected from breaking your use case in the future.

Thanks for reporting and fixing this issue, much appreciated.

sirpercival:
OK, I added it. I haven't run the tests, because I was using the pip version of jsonpickle and not the git version, but hopefully that should work fine.

Issue #135, opened 2015-10-21
Problem with list of Enums


Umqra:
This simple code cause `IndexError: list index out of range` ([full stack-trace](http://pastebin.com/D450xBUt))
```python
import jsonpickle
import enum

class TestEnum(enum.Enum):
    Test = 1

s = [TestEnum.Test, TestEnum.Test]
dumps = jsonpickle.dumps(s)
loads = jsonpickle.loads(dumps)
```

Jsonpickle dumps first `TestEnum.Test` to `{"py/reduce": ...}` and second `TestEnum.Test` to `{"py/id": 1}`. But when jsonpickle decode first enum, it doesn't create reference to this object and it causes `IndexError` while parsing second enum.

It seems, at the moment jsonpickle can't serialize this list... Or am i wrong?

Anna-Bonus:
Hello, @Umqra

You are not wrong. It's a bug in jsonpickle - it doesn't support references in reduced objects. To fix it just add follow line to the end of Unpickler._restore_reduce in unpickler.py: 

```
--> self._mkref(stage1)
return stage1
```

Link to the line in source code: https://github.com/jsonpickle/jsonpickle/blob/master/jsonpickle/unpickler.py#L207

@davvid can you fix it in master branch?

davvid:
Thanks for the report @Umqra , and thanks for the PR @Anna-Bonus 

Please test this out -- it should be fixed now in master.

PR #137, opened 2015-11-14
Fix bug with references to objects which support __reduce__ interface

Anna-Bonus:
Particularly it's a solution for #135


davvid:
Thanks much @Anna-Bonus

PR #145, opened 2016-05-05
ndarray views and binary encoding

EelcoHoogendoorn:
This PR adds support for ndarray views, or tracking references inside nd-arrays. In a nutshell:

```
        # the fact that this works is a defining feature of this package
        row = [1, 2, 3]
        nested_list = [row, row]
        nested_list_ = roundtrip(nested_list)
        nested_list_[0][0] = -1
        assert nested_list_[1][0] == -1

        # so it would be kinda nice if this worked the same way; which it does as per this PR
        row = np.arange(3)
        jagged_array = [row, row[:]]
        jagged_array_ = roundtrip(jagged_array)
        jagged_array_[0][0] = -1
        assert jagged_array_[1][0] == -1
```

Also, arrays with more elements than a given threshold are encoded as (optionally) compressed b64. This is a much simpler and somewhat separate issue; so perhaps it does not belong in this PR, though personally I would really like to see that merged as well. For big arrays, the computational overhead of conversions to string are huge, and the human-readability argument for them is pretty much void.

Feedback is very much welcome!


EelcoHoogendoorn:
Also, this PR fixes a bug when pickling arrays with a shape containing zeros; see the test suite.

tests/numpy_test.py
```diff
@@ -106,6 +109,146 @@ def test_ndarray_roundtrip(self):
...
+    def test_views(self):
+        """Test that views are maintained under serialization"""
+        rng = np.arange(20)  # a range of an array
+        view = rng[10:]  # a view referencing a portion of an array
+        data = [rng, view]
+
+        _data = self.roundtrip(data)
+
+        _data[0][15] = -1
+        assert _data[1][5] == -1
```
davvid:
```python
self.assertEqual(-1, _data[i][5])
```
perhaps?

tests/numpy_test.py
```diff
@@ -106,6 +109,146 @@ def test_ndarray_roundtrip(self):
...
+
+    def test_immutable(self):
+        """test that immutability flag is copied correctly"""
+
+    # def test_newstrided(self):
+    #     a = np.arange(10)
+    #     q = np.ndarray(buffer=a.data, strides=a.strides*2, shape=(5,6), dtype=a.dtype)
+    #     assert q.base is a
+    #     print(q)
+
```
davvid: If we're going to have a commented-out test is it worth having a TODO comment explaining how we might handle it in the future?  Otherwise, a short comment explaining why it doesn't work(?) might help.  Ditto with the empty `test_immutable` above.

tests/numpy_test.py
davvid: Ditto for the rest of the asserts below. The python `self.assertEqual(...)`, `self.assertTrue(...)` family of functions are preferred over `assert` because the latter is optimized away when running python in optimized mode, and generally the suggested way to write test code.

.gitignore
```diff
@@ -10,3 +10,4 @@
 /tags
 __pycache__
 .tox/
+.idea
```
davvid: Please avoid no-newline-at-end-of-file. If you're using sublime text please configure it to not do that[1].

It causes merge churn.  It might be worth rebasing to fix that and squash some of these later fixes into their original commit in this series too.

[1] https://forum.sublimetext.com/t/make-saving-newline-at-eof-the-installation-default/9842

jsonpickle/ext/numpy.py
```diff
@@ -1,6 +1,11 @@
 # -*- coding: utf-8 -*-
 
 from __future__ import absolute_import
+from builtins import *
```
davvid: Have you run `make tox`?   I don't think there's a `builtins` module in python2.

```sh
$ python -V
Python 2.7.11

$ python -mbuiltins
/usr/bin/python: No module named builtins
```

Also, it's probably worth explicitly mentioning each symbol we want to import. We avoid blind `import *` for many reasons.  There should be a `make check` target you can use to run `flake8` on the code base, which should have flagged this line.

jsonpickle/ext/numpy.py
```diff
@@ -58,18 +65,175 @@ class NumpyNDArrayHandler(NumpyBaseHandler):
...
+
+
+        return arr
 
 
 def register_handlers():
     jsonpickle.handlers.register(np.dtype, NumpyDTypeHandler, base=True)
     jsonpickle.handlers.register(np.generic, NumpyGenericHandler, base=True)
-    jsonpickle.handlers.register(np.ndarray, NumpyNDArrayHandler, base=True)
+    jsonpickle.handlers.register(np.ndarray, NumpyNDArrayHandlerView, base=True)
```
davvid: Just curious, does this break backwards-compatibility with an existing json file that serialized with the older handler? If so we'll probably want to add a compatibility warning and bump the minor version up on next release.  Hopefully this handler can take the old data too?

jsonpickle/ext/numpy.py
```diff
@@ -58,18 +65,175 @@ class NumpyNDArrayHandler(NumpyBaseHandler):
...
+
+    'base' arrays, or arrays which are viewed by other arrays, must be f-or-c-contiguous.
+    This is not such a large restriction in practice, because all numpy array creation is c-contiguous by default.
+    Releasing this restriction would be nice though; especially if it can be done without bloating the design too much.
+
+    Furthermore, ndarrays which are views of array-like objects implementing __array_interface__,
+    but which are not themselves nd-arrays, are deepcopied with a warning,
+    as we cannot guarantee whatever custom logic such classes implement is correctly reproduced.
+    """
+    mode = 'warn'
```
davvid: Should we document the end-user interface to the `warn` variable?  I can imagine someone might want to set it to `silent` for example, so it might be worth showing how the user might do that.

davvid:
Done reviewing. Other than those minor notes this looks really great.  Looking forward to merging this!

tests/numpy_test.py
EelcoHoogendoorn: Thanks for the suggestion, I will implement that

tests/numpy_test.py
EelcoHoogendoorn: I thought of some improvements after I made this PR, so some things are still in flux around the edges. But certainly, I will clean this kind of thing up before i am done.

EelcoHoogendoorn:
One open question I still have is how to make the serialization behavior more composable. There are several parameters to the current handlers; when to store as binary, if ever; what compression to apply, if any; when to store views and when to make deep copies; and maybe others.

Seems like it would not be too hard to allow constructor arguments on handler subclasses, and register an instance instead of the class; just need a little refactor to bind the context differently depending on if the registered handle is a class or instance. Thoughts?

davvid:
Constructor arguments would be a good general-purpose way of fixing it.  Another way would be to have the pickler/unpickler have a `self.handler_data = {}` that handlers can use for stashing things via `self.context.handler_data['numpy.foo'] = 'bar'` but I'm not sure if that'd be a little too much of a coupling between handler and pickler/unpickler.

jsonpickle/ext/numpy.py
EelcoHoogendoorn: not actually used, so removed it

jsonpickle/ext/numpy.py
EelcoHoogendoorn: yes, it should be backwards compatible; though it never hurts to test this explicitly

jsonpickle/ext/numpy.py
EelcoHoogendoorn: yeah... this relates to my comment below; would be nice if the handlers had an init method

davvid: I guess the choice is between allowing to register instances, or adding a mechanism for specifying constructor args to specific handlers.  It does seem like allowing for instances would be the simplest and most versatile design without needing to do too much plumbing for handler-specific arguments.

EelcoHoogendoorn: good one!

EelcoHoogendoorn:
Yeah; another thing is the inheritance relation I have right now does not allow you to

EelcoHoogendoorn:
Thanks for the feedback! Id like to add some tests relating to byte-ordering still (not 100% sure that works correctly with views yet), and add init methods to the handlers as discussed; but I think that will be all for now then.

EelcoHoogendoorn:
By the way; id recommend bumping the minor version number; even though I expect existing data to load just fine, people may be relying on the old behavior, where serialization destroys views.

EelcoHoogendoorn:
Good to go, as far as I am concerned!

EelcoHoogendoorn:
I think the python version compatibility issues have been resolved; but it would be nice to add CI I suppose. only tested on windows so far.

Ive reworked the way byteorder information is stored quite a bit. The byteorder of views is now stored relative to their base array (absent in the default case of identical byteorder), which also means that there is no need to store it with all text arrays, which was kinda verbose and ugly.

jsonpickle/pickler.py
```diff
@@ -251,7 +251,7 @@ def _flatten_bytestring(self, obj):
                 return obj.decode('utf-8')
             except:
                 pass
-        return {tags.B64: base64.encodestring(obj).decode('utf-8')}
+        return {tags.B64: base64.encodebytes(obj).decode('utf-8')}
```
davvid: `base64.encodebytes()` does not exist in python2.  Is the end result different?  My naive testing showed that they're interchangeable, but I may be missing something.

I think you should be able to turn on the travis CI support on your clone and it should run the tests on both python2 and python3 in case that helps run the python2 tests without needing to install it locally.

Here are a few of the test failures when running on python2.7

```
ERROR: test handling of non-c-contiguous memory layout
  File "tests/numpy_test.py", line 181, in test_transpose
    _a, _b = self.roundtrip([a, b])
  [...]
  File "jsonpickle/pickler.py", line 284, in _flatten_obj_instance
    return handler(self).flatten(obj, data)
  File "jsonpickle/ext/numpy.py", line 221, in flatten
    if not obj.data.c_contiguous:
AttributeError: 'buffer' object has no attribute 'c_contiguous'
```

This failure seems a little surprising. The same test passes in a python3.4 virtualenv. Same line of code, different error message.
```
ERROR: test that cases with non-standard strides and offsets work correctly
[...]
    return handler(self).flatten(obj, data)
  File "jsonpickle/ext/numpy.py", line 221, in flatten
    if not obj.data.c_contiguous:
AttributeError: cannot get single-segment buffer for discontiguous array
```

A couple of the tests also run into `AttributeError: 'module' object has no attribute 'encodebytes'` as mentioned above.

This last one is also surprising, though maybe not so much because the test *is* called "test_weird_arrays", so perhaps it's unsurprising? ;-)

```
FAIL: test behavior with memoryviews which are not ndarrays
  File "tests/numpy_test.py", line 224, in test_buffer
    self.assertEqual(len(w), 1)
AssertionError: 0 != 1
  
FAIL: test that we disallow serialization of references to arrays that do not effectively own their memory
  File "tests/numpy_test.py", line 165, in test_weird_arrays
    self.assertEqual(len(w), 1)
AssertionError: 0 != 1
```

EelcoHoogendoorn:
To add to my previous comment; I think it would be ideal if the serialization behavior of both the stride order and byte order would be configurable. From a serialization-purity perspective, we would like to restore the exact byteorder and strides that we put in in the first place; but for practical intents and purposes, it is typically much more desirable to take the opportunity to 'normalize' these attributes to their numpy/platform defaults while we have the chance. There are arguments to be made for both, and I don't think this choice should be forced at the library level. Would be nice if in the constructor we could specify byteorder and strideorder as either relative to platform default, or absolute. But it will require some tinkering.

Another question; should the binary format be explicitly stored? Already ran into this problem when applying jsonpickle, when changing from bz2 to zlib compression... would also make it easier to support npz format as well.

By the way, I realize I havnt really given any thought yet to object arrays; they are fundamentally rather different from a serialization perspective, since they contain references. I suppose these should always be serialized as text... which is fine since they are rarely very big anyway... Ill make some tests for it.

EelcoHoogendoorn:
```
print (jsonpickle.encode([dict(a=1)]*2, make_refs=True))
[{"a": 1}, {"a": 1}]
```
Ran into this while working on object arrays; is this a bug? I would expect to see one dict and a reference here. Or are references to primitive types such as lists and dicts not supported?

davvid:
For simplicity we only track references to objects that can't be ordinarily encoded by json as-is. We don't track dicts or lists but I can see how that could be useful, and it is a better, more general solution.

The referencing machinery is something that could benefit from a redesign.  Right now the IDs stored in json are basically the index into the list of objects we've seen as traversing, which can be fragile and probably would break down if we were to properly handle list and dict references.

A better way might be to change the storage into a flat object dict keyed by IDs (`id(...)` like we currently do?) that we store and restore in dependency (and reverse) order.  In order to ensure that we can recreate all references we could also store the list of keys so that the decoder can traverse the objects in the reverse order as it did when encoding them.

If you `git grep _Proxy` you'll see one part of the current design that would be really great to eliminate/redo.  Another possibly questionable thing is that we're dependent on the final json being written with sorted keys (`git grep sort`).  It would be nice if that wasn't required.

davvid:
If explicitly storing the binary format helps the current (and possible future) formats then that sounds like a good idea to include IMO.

EelcoHoogendoorn:
Thanks for the elaboration; looking forward to the redesign that unifies this behavior. This behavior surprised me as well; is there a rationale for that?

```
print (jsonpickle.encode([[NumpyTestCase]]*2, make_refs=True))
[[{"py/type": "__main__.NumpyTestCase"}], {"py/id": 1}]
print (jsonpickle.encode([NumpyTestCase]*2, make_refs=True))
[{"py/type": "__main__.NumpyTestCase"}, {"py/type": "__main__.NumpyTestCase"}]
```

Wrt the sorting: from a human readability perspective, I think itd be nice if all objects being referenced explicitly stored their own unique id, rather than having this id encoded implicitly. That way, if I look at my json and see a py/id tag, I could quickly crtl-f the corresponding id to jump there.

Korijn:
What work still needs to be done on this PR to get it merged?

davvid:
@Korijn the PR as-is breaks python2 so we can't merge it until those issues are addressed. That said, I see there was a `last 2.7 fix` commit, which should help, but I don't know if the issues about "no newline at end of file" were ever addressed.

I still see `base64.encodebytes` in the diff, so I think that change needs to be removed since it breaks python2 and it doesn't change any observable behavior for other versions, so it doesn't make sense to change that.  I think if that's addressed then it should be python2-compatible.

I just double-checked and `make tox` still fails for python2.7.  It does pass for python3.4.  One more fix, pretty please?

EelcoHoogendoorn:
Sorry for the delay; my list of things that id ideally still like to change to this PR kept growing, while my amount of free time kept shrinking. Finally have my new apartment more or less in order though, so hopefully one of these days I can polish this off so it is good to go.

PR #159, opened 2017-01-10
fixed serialization of fortran ordered base arrays

EelcoHoogendoorn:
this had a critical bug due to an incorrect numpy docstring

Id love to see a patch / release for this asap, as this is a really bad bug


davvid:
Thanks!

PR #167, opened 2017-03-27
[SimpleReduceHandler] handle more than two args

### LINKED ISSUE #163: Unpickling collection.deque fails with ValueError on Python 3.6
```
ERROR: Make sure we can handle collections.deque
  File "/private/tmp/jsonpickle/jsonpickle/tests/object_test.py", line 315, in test_deque_roundtrip
    new_deque = jsonpickle.decode(encoded)
  File "/private/tmp/jsonpickle/jsonpickle/jsonpickle/__init__.py", line 152, in decode
    return unpickler.decode(string, backend=backend, keys=keys)
  File "/private/tmp/jsonpickle/jsonpickle/jsonpickle/unpickler.py", line 27, in decode
    return context.restore(backend.decode(string), reset=reset)
  File "/private/tmp/jsonpickle/jsonpickle/jsonpickle/unpickler.py", line 120, in restore
    value = self._restore(obj)
  File "/private/tmp/jsonpickle/jsonpickle/jsonpickle/unpickler.py", line 162, in _restore
    return restore(obj)
  File "/private/tmp/jsonpickle/jsonpickle/jsonpickle/unpickler.py", line 243, in _restore_object
    instance = handler(self).restore(obj)
  File "/private/tmp/jsonpickle/jsonpickle/jsonpickle/handlers.py", line 218, in restore
    factory, args = [restore(i, reset=False) for i in data['__reduce__']]
ValueError: too many values to unpack (expected 2)

```

marscher:
Fixes #163

https://travis-ci.org/marscher/jsonpickle/builds/215615881

### DIFF
--- a/jsonpickle/handlers.py
+++ b/jsonpickle/handlers.py
@@ -215,8 +215,24 @@ class SimpleReduceHandler(BaseHandler):
 
     def restore(self, data):
         restore = self.context.restore
-        factory, args = [restore(i, reset=False) for i in data['__reduce__']]
-        return factory(*args)
+        res = [restore(i, reset=False) for i in data['__reduce__']]
+        factory = res.pop(0)
+        args = res.pop(0)
+        obj = factory(*args)
+
+        if len(res) >= 1:
+            state = res[0]
+            if state and hasattr(obj, '__setstate__'):
+                obj.__setstate__(state)
+            if len(res) >= 2:
+                iterator = res[1]
+                obj.extend(iterator)
+            if len(res) == 3:
+                item_iter = res[2]
+                for k, v in item_iter:
+                    obj[k] = v
+
+        return obj
 
 
 class OrderedDictReduceHandler(SimpleReduceHandler):
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,7 @@ matrix:
   include: 
     - env: TOXENV=py27
       python: 2.7
-      
-    - env: TOXENV=py33
-      python: 3.3
-      
+     
     - env: TOXENV=py34
       python: 3.4
       
@@ -15,9 +12,6 @@ matrix:
       python: 3.5
       
     - env: TOXENV=py36
-      python: 3.6
-
-  allow_failures:
       python: 3.6
 
 install: 


jsonpickle/handlers.py
```diff
@@ -215,8 +215,24 @@ def flatten(self, obj, data):
...
     def restore(self, data):
         restore = self.context.restore
-        factory, args = [restore(i, reset=False) for i in data['__reduce__']]
-        return factory(*args)
+        res = [restore(i, reset=False) for i in data['__reduce__']]
+        factory = res.pop(0)
+        args = res.pop(0)
+        obj = factory(*args)
+
+        if len(res) >= 1:
```
marcintustin: You might as well keep popping until res is empty, or use izip_longest, but fundamentally this is a good catch.

marscher: Thanks for the suggestions. I thought izip_longest is used to join two sequences with a fill value, if one sequence length exceeds the other one. If one pops the list until its empty, it is more difficult to assign the values properly.

davvid: I've merged this version as-is but please do submit another PR if you think we can improve this part even further.  Thanks for the careful review @marcintustin.

marcintustin:
So, I now remember why this was weird to me - jsonpickle's  default Pickler and Unpickler classes already handle the full reduce protocol correctly. This just introduces code duplication. I'm going to look at removing this class other than as a wrapper. 

Either that, or we should return this as it was, to explicitly not handle the full protocol. While that might sound crazy, it makes sense if we also deprecate it and put it on a path for removal. @davvid What are your thoughts.

marcintustin:
Specifically, it looks like the problem that this handler was solving is that certain classes are not getting serialised using the reduce method

davvid:
Removing the duplication by making the class a wrapper would be a nice improvement, though I'll take any improvement to how it's currently implemented.  It's true that the handler is there to catch those other classes; if there's a more general approach that simplifies the code then I'm all for it.

marscher:
The thing is, that the default pickler also invokes getstate even if the reduce protocol is present. So in a class hierarchy with getstate present, one can not give the reduce protocol the highest priority. Note that getstate is ignored by Python (2 and 3), if the reduce protocol is available.

PR #168, opened 2017-03-30
made internal imports relative for the sake of easy submodule inclusion.

marscher:
The rationale for this change is the following. If somebody wants to include jsonpickle within a repository and directly import from this new location (eg. pkg_foo.libs.jsonpickle), this is not yet possible, because all internal import are absolute to the "jsonpickle" package.

I know that there are concerns using relative imports, but it is actually the only way to make the package relocatable within another package.

https://stackoverflow.com/questions/4209641/absolute-vs-explicit-relative-import-of-python-module

### DIFF
--- a/jsonpickle/__init__.py
+++ b/jsonpickle/__init__.py
@@ -53,13 +53,13 @@ added to JSON::
     assert obj.name == result['name'] == 'Awesome'
 
 """
-from jsonpickle import pickler
-from jsonpickle import unpickler
-from jsonpickle.backend import JSONBackend
-from jsonpickle.version import VERSION
+from . import pickler
+from . import unpickler
+from .backend import JSONBackend
+from .version import VERSION
 
 # ensure built-in handlers are loaded
-__import__('jsonpickle.handlers')
+from . import handlers as _
 
 __all__ = ('encode', 'decode')
 __version__ = VERSION
--- a/jsonpickle/backend.py
+++ b/jsonpickle/backend.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 
-from jsonpickle.compat import PY32
-from jsonpickle.compat import unicode
+from .compat import PY32
+from .compat import unicode
 
 
 class JSONBackend(object):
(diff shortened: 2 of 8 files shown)


jsonpickle/__init__.py
```diff
@@ -53,13 +53,13 @@ def __init__(self, name):
     assert obj.name == result['name'] == 'Awesome'
```
davvid: This is a nice improvement.  It might make sense to also add from-future imports while we're in here:

```python
from __future__ import absolute_import, division, unicode_literals
```

marcintustin:
The tests fail for me locally with 13 errors like the below:

```
ERROR: Failure: ImportError (cannot import name util)
  File "/Users/marcin/Documents/code/virtualenvs/jsonpickle/lib/python2.7/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/Users/marcin/Documents/code/virtualenvs/jsonpickle/lib/python2.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/Users/marcin/Documents/code/virtualenvs/jsonpickle/lib/python2.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/Users/marcin/Documents/code/jsonpickle/tests/util_test.py", line 13, in <module>
    import jsonpickle.util
  File "/Users/marcin/Documents/code/jsonpickle/jsonpickle/__init__.py", line 58, in <module>
    from . import pickler
  File "/Users/marcin/Documents/code/jsonpickle/jsonpickle/pickler.py", line 15, in <module>
    from . import util
ImportError: cannot import name util
```

That happens when invoking with `nosetests --with-doctest jsonpickle tests/` or tox. 

Also the instructions on running the test suite are so out of date they refer to files that don't exist: http://jsonpickle.github.io/contrib.html#run-the-test-suite

marcintustin:
Ignore my previous comment, it was caused by something I was doing

marcintustin:
Nevertheless, the test instructions are out of date: http://jsonpickle.github.io/contrib.html#run-the-test-suite

Issue #224, opened 2018-08-31
translate/untranslate_module_name should prefer Python 3 naming


jaraco:
Looking at the [util module](https://github.com/jsonpickle/jsonpickle/blob/e42147fb39da16557f2e73d89bce582e041dd54c/jsonpickle/util.py#L450-L482), I notice that the functions prefer the Python 2 names to the Python 3 names.

I made similar poor choices early in my early years of Python 3 porting, but if I've learned anything from my experience, it's that compatibility code should always aim to use the future constructs first and provide backward compatibility to the older constructs. Otherwise, you end up in a state where you have legacy constructs in newer platforms indefinitely. In this case, Python 3-only applications will still be storing pickles using Python 2 names.

At some point, this package is going to want to shed that legacy. I suggest sooner than later.

PR #225, opened 2018-08-31
Prefer 'builtins' module name

jaraco:
As laid out in #224, futurize the code to prefer the more modern naming.

I'd particularly like scrutiny on 8383649 as it removes what appears to be unused code. The tests continue to pass with that removal, but I worry there's untested behavior relying on that removed code. Can you think of any example that would be affected by that change?

### DIFF
--- a/jsonpickle/util.py
+++ b/jsonpickle/util.py
@@ -441,22 +441,22 @@ def has_reduce(obj):
 
 
 def translate_module_name(module):
-    """Rename builtin modules to a consistent (Python2) module name
+    """Rename builtin modules to a consistent module name.
+
+    Prefer the more modern naming.
 
     This is used so that references to Python's `builtins` module can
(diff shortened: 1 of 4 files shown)


jsonpickle/util.py
```diff
@@ -243,21 +243,14 @@ def is_function(obj):
...
-                     'instancemethod',
-                     'method-wrapper'))
+    function_types = (
+        types.FunctionType,
+        types.MethodType,
+        types.LambdaType,
+        types.BuiltinFunctionType,
+        types.BuiltinMethodType,
+    )
+    return type(obj) in function_types
```
davvid: This seems like it should do the right thing.  Would it be simpler still as `isinstance(obj, function_types)`?  Merged as-is nonetheless.  If we discover anything that relies on the original implementation we'll add a test then.  I suspect these weren't available in python 2.5 or 2.6 back when it was written.

PR #251, opened 2019-03-02
Add Base85 encoding support for Python 3

### LINKED ISSUE #250: Base-85 encode bytes on Python 3.4+?
The `base64` library supports [base-85 encoding](https://docs.python.org/3/library/base64.html#base64.b85encode) as of Python 3.4. This is a more efficient form of text encoding, that bloats binary data by only 1/4 as opposed to base-64 which expands binary data by 1/3. From some manual testing I've done with randomized bytes given by `os.urandom()`, this change makes payloads about 10% smaller.

Given that Python 3.3 has reached end of life and is no longer supported by pip, and Python 2 can't read data pickled by Python 3 anyway, I think it'd be cool if we could add an option to `encode()` to encode binary data using base-85.

As long as we provide this as an additional argument to `encode()`, this would be backwards-compatible and shouldn't break anything (I believe). What do people think?

dargueta:
TL;DR: Allow using [Base85](https://en.wikipedia.org/wiki/Ascii85) text encoding on Python 3.

## Details

This doesn't remove base64 encoding support for Python 3; rather, it makes base85 the default but still allows overriding the default to only use base64 for backwards compatibility with Python 2. (Python 2 doesn't support base85, nor does there appear to be a backport for it.)

## Why
It's a more compact text encoding than Base64, and some manual experimentation gives about a 10% space savings for randomized binary data.

Closes #250.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -9,6 +9,15 @@ Version 1.2 - TBD 2019
       These objects are now handled robustly.
       (`#247 <https://github.com/jsonpickle/jsonpickle/issues/247>`_).
 
+    * `Base85`_ is now the default binary data encoding on Python 3, as it's
+      more space-efficient than base64. Python 3 can still read and write
+      base64-encoded data, but Python 2 cannot read nor write base85. For
+      backwards compatibility, the pickler can be configured to encode using
+      only base64 via the new ``prefer_base85`` argument.
+      (`#251 <https://github.com/jsonpickle/jsonpickle/issues/251>`_).
+
+.. _Base85: https://en.wikipedia.org/wiki/Ascii85
+
 Version 1.1 - January 22, 2019
 ------------------------------
     * Python 3.7 `collections.Iterator` deprecation warnings have been fixed.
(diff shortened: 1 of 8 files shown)


davvid:
Sweet, this looks perfect except for one little trivial tweak that I'll fixup here before pushing out the merge. I'm going to rename the variables from `prefer_base85` to `use_base85`, for consistency with the existing `use_decimal`.

cyberic99:
Hi,

Has anyone done any perf mesurement after switching to base85?

dargueta:
Oh, good point. All of my testing has been on relatively simple stuff.

davvid:
We haven't tagged a release yet so changing this default is still fair game.

dargueta:
I looked at the Python 3.6 source code and it appears that there's significant overhead on _the first_ call to b85encode because it builds some tables, but subsequent calls are faster. Still, it's noticeably slower than base64.

I'll open a PR to switch base64 to be the default later today. Thanks @cyberic99 for pointing out something I obviously should've done first. Sorry about this!

dargueta:
Done: https://github.com/jsonpickle/jsonpickle/pull/252

cyberic99:
Thx @dargueta !

For my use case, the thoughput is more important than the file size, so I thought about the performance impact right away ;-)

Issue #239, opened 2018-11-12
Error encoding numpy scalars in Python 2.7


prabhuramachandran:
On Python 2.7.13 with jsonpickle 1.0 and numpy 1.13.3, I run into an error with the following:
```
import jsonpickle
import numpy as np
x = np.linspace(0, 1)
jsonpickle.loads(jsonpickle.dumps(x[0]))
```
This fails with a `TypeError` (am happy to upload a full traceback).  It works fine on Python 3.6 and numpy 1.14.1.

AndersonReyes:
@prabhuramachandran you have to register the numpy extension

```
import jsonpickle.ext.numpy as jpnp
jpnp.register_handlers()
```

prabhuramachandran:
Yes that seems to have fixed it.  I am curious why that was not needed on Python 3.6?  That is what tripped me up and made me think this was a bug.

AndersonReyes:
that is a good question hmmmm, seems to be happening at that stage1.__setstate__(state)

PR #256, opened 2019-04-04
Further Pandas Support

parsons-kyle-89:
The pandas extension library fails under certain use cases.  In particular, non-standard indexes on data frames will not serialize/unserialize correctly.  In certain cases, an error is even thrown upon serialization.  This PR aims to handle pandas objects better in the pandas extension library.

The main thrust of this PR is that a `pandas.DataFrame` will now be saved with its index not appearing in the `value` portion of the serialization. Instead the index is saved entirely separately in the `index` key of the `meta` portion of the serialization. Further, additional handlers have been added for several different types of indexes (though most of the code between them is reused). Handlers were also added for timestamps, periods, and intervals.

For more context of what doesn't work in the current handler:
```
import pandas as pd
from pandas.testing import assert_frame_equal

import jsonpickle as jp
import jsonpickle.ext.pandas
jp.ext.pandas.register_handlers()

def roundtrip(obj):
    return jp.decode(jp.encode(obj))

df = pd.DataFrame({"a": [1, 2], "b": ["foo", "bar"]})
assert_frame_equal(df, roundtrip(df)) # this is fine

idx = pd.Index(pd.date_range('2019-01-01', '2019-01-02'))
df.set_index(idx, inplace=True)
assert_frame_equal(df, rountrip(df)) # this assertion fails because the index of df comes back as object type instead of DatetimeIndex type

idx = pd.IntervalIndex.from_breaks(range(3))
df.set_index(idx, inplace=True)
jp.encode(df) # this already fails seemingly due to a bug with pd.DataFrame.to_csv
```

### DIFF
--- a/jsonpickle/ext/pandas.py
+++ b/jsonpickle/ext/pandas.py
@@ -8,6 +8,7 @@ from .. import encode, decode
 from ..handlers import BaseHandler, register, unregister
 from ..util import b64decode, b64encode
 from ..backend import json
+from .numpy import register_handlers as register_numpy_handlers, unregister_handlers as unregister_numpy_handlers
 
 __all__ = ['register_handlers', 'unregister_handlers']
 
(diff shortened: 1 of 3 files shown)


jsonpickle/ext/pandas.py
```diff
@@ -120,27 +116,102 @@ def restore(self, data):
...
+        _, meta = self.pp.restore_pandas(data)
+        left = decode(meta['left'])
+        right = decode(meta['right'])
+        closed = str(meta['closed'])
+        obj = pd.Interval(left, right, closed=closed)
+        return obj
+
+
 def register_handlers():
+    register_numpy_handlers()
```
parsons-kyle-89: This seems not ideal, but I was running into some strange errors when running the tests vs python 2 that this fixed. It seems related to [this](https://github.com/jsonpickle/jsonpickle/issues/239) since the errors only occurred in python 2 and not python 3.

jsonpickle/ext/pandas.py
```diff
@@ -120,27 +116,102 @@ def restore(self, data):
...
 
+    index_constructor = pd.Index
+    name_bundler = lambda _, obj: {'name': obj.name}
+
     def flatten(self, obj, data):
-        meta = {'dtype': str(obj.dtype), 'name': obj.name}
-        buf = json.dumps(obj.tolist())
+        name_bundle = self.name_bundler(obj)
+        meta = dict(dtype= str(obj.dtype), **name_bundle)
+        buf = encode(obj.tolist())
```
parsons-kyle-89: This is an important difference.  The data of an index is now `encod`ed instead of just `json.dumps`.  This allows for the handling of more complex pandas types in indexes.

davvid: Thanks for the comments explaining that bit, this looks great.

PR #260, opened 2019-04-20
fix #259 prevent stackoverflow in _getstate

### LINKED ISSUE #259: foxdot breaks jsonpickle
see https://github.com/Almenon/AREPL-vscode/issues/205

can be replicated with simply:

```python
from FoxDot import pluck

import jsonpickle

foo = jsonpickle.encode(pluck)
```

Almenon:
see https://github.com/jsonpickle/jsonpickle/issues/259

flatten will increase the depth, thus preventing a stackoverflow.

### DIFF
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -593,7 +593,7 @@ class Pickler(object):
                                       make_refs=self.make_refs)
 
     def _getstate(self, obj, data):
-        state = self._flatten_obj(obj)
+        state = self._flatten(obj)
         if self.unpicklable:
             data[tags.STATE] = state
         else:
--- a/tests/object_test.py
+++ b/tests/object_test.py
@@ -76,6 +76,10 @@ class GetstateReturnsList(object):
     def __setstate__(self, state):
         self.x, self.y = state[0], state[1]
 
+class GetstateRecursesInfintely(object):
+    def __getstate__(self):
+        return GetstateRecursesInfintely()
+
 
 class ListSubclassWithInit(list):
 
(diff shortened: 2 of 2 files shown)


Almenon:
two problems; vx doesn't have a windows script, and when installing dependencies I got this error:

```
Collecting yajl (from -r requirements-dev.txt (line 16))
  Downloading https://files.pythonhosted.org/packages/70/5a/57ed8b430a55e2bfc9c4ea0a263f41f4d2fe5d962ff073c8af117da78be1/yajl-0.3.5.tar.gz (52kB)
     |████████████████████████████████| 61kB 4.1MB/s
    ERROR: Complete output from command python setup.py egg_info:
    ERROR: fatal: not a git repository (or any of the parent directories): .git
    >>> Creating a symlink for compilationg: includes/yajl -> yajl/src/api
      File "<string>", line 1, in <module>
      File "C:\Users\almenon\AppData\Local\Temp\pip-install-nxe3h36b\yajl\setup.py", line 73, in <module>
        os.symlink(os.path.join('..', 'yajl', 'src', 'api'), os.path.join('includes', 'yajl'))
    OSError: symbolic link privilege not held
    ----------------------------------------
ERROR: Command "python setup.py egg_info" failed with error code 1 in C:\Users\almenon\AppData\Local\Temp\pip-install-nxe3h36b\yajl\
```

Luckily I have WSL and it worked in there. Now I have to add the test but I have to work soon so guess that will be for later.

davvid:
It looks like `yajl` doesn't build on windows because of the symlink kerfufle.  Delete that line from `requirements.txt` and you should be good to go.

I'm going to try to fix it by [checking the platform.](https://stackoverflow.com/questions/29222269/is-there-a-way-to-have-a-conditional-requirements-txt-file-for-my-python-applica/35614580#35614580)

What does `sys.platform` look like on the one that fails to compile, `win32`?  I'll go with that for now so you might be able to rebase against a `requirements.txt` fix shortly.  If it's something else let me know or include it in the PR.  Thanks again

Almenon:
yep, win32.

Almenon:
done :)

davvid:
Sweet, thanks!

Issue #241, opened 2018-12-01
Option to remove 'null' fields


turian:
For some classes, it can be much more concise to ignore/drop any field that is `null`. Could there be an optional parameter that enables this?

valtron:
At the very least, this would be nice for `py/reduce`: the [last three fields are optional](https://docs.python.org/3/library/pickle.html#object.__reduce__) anyway.

ZENALC:
Not sure if it helps, but here is a brute-force method of doing it. In pickler, find the encode function and change the return backend.encode(context.flatten(value, reset=reset), indent=indent, separators=separators) with the code below.

> 
    finalResult = backend.encode(context.flatten(value, reset=reset), indent=indent, separators=separators)
    if nullValues:
        return finalResult
    else:
        previousResult = finalResult
        finalResult = finalResult.replace("null", "None")
        evaluatedResult = eval(finalResult)
        if type(evaluatedResult) is not dict:
            return previousResult
        newDict = {}
        for k, v in evaluatedResult.items():
            if v is not None:
                newDict[k] = v
        return str(newDict)

PR #261, opened 2019-05-01
don't emit null optional arguments for `py/reduce`

valtron:
The last three elements of [`__reduce__`](https://docs.python.org/3/library/pickle.html#object.__reduce__) are optional and are typically `None`. This PR omits trailing `null`s in `py/reduce` to make the JSON slightly nicer.

Related to #241.


davvid:
The implementation is pretty brilliant.  Thanks for the enhancement.

PR #263, opened 2019-05-21
ujson: disable escape_forward_slashes by default

rylz:
otherwise ujson's output is incompatible with other jsonpickle backends
because it escapes the slashes in "py/object" etc.

also moved encoder options defaults to after all of the load_backend
calls, because previously those calls were overwriting the
encoder_options with `((), {})` for every encoder


davvid:
Thanks, yeah, that's one of the worst defaults ever.  Who the heck likes escaped slashes?

PR #268, opened 2019-06-25
Fix: restoring an empty Pandas DF fails

misantroop:
read_csv cannot read a file w/o columns. Check if file to restore from is empty (\r\n), create new dataframe if needed.


davvid:
Makes sense, this looks like a sensible fallback.  Thanks!

Issue #205, opened 2018-05-18
[Wishlist] fail-safe mode


Almenon:
for [AREPL](https://github.com/Almenon/AREPL) I use jsonpickle to encode the user's variables after exec() for later display to the user.

This works great the vast majority of the time.  However, there are some odd instances in which jsonpickle fails.  For example, https://github.com/Almenon/AREPL-backend/issues/32 .  In these cases rather than failing completly I would prefer the pickling to continue, leaving a message in place of the unpickleable variable.

If I have time I might try implementing this myself and submitting a pull request.

Almenon:
I implemented this and it's working well in https://github.com/Almenon/AREPL-vscode - I need to submit a PR.

PR #269, opened 2019-06-30
Fail safe mode

Almenon:
see #205

re-created PR because previous PR's history got wonky.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,15 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Python: Current File",
+            "type": "python",
+            "request": "launch",
+            "program": "${file}",
+            "console": "integratedTerminal"
+        }
+    ]
+}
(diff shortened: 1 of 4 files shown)


davvid:
Thanks.  In the future please squash commits like "fix bad rebase" into the original commit that introduced the issue so that we don't have cruft in the permanent history.

Heads-up -- I deleted the `.vscode` directory.  I don't want to have platform-specific editor settings in the project -- you should be able to do `echo /.vscode >> .git/info/exclude` to have git ignore the directory in your worktree only.  Maybe we could have an "editor config" file in here, but I'd prefer not to because our style is 100% PEP-8 and it's weird to carry around cruft just for vscode.  Taken to the logical extreme, we would need to maintain configs for all kinds of editors, across all projects I maintain, which does not scale.  Thanks for understanding, and thanks for fixing up the history through your initial rebase.

The implementation was nice and clean -- merged!  Thanks for adding a test case too -- that'll keep me from breaking it in the future.  Please take a look at b4b4333f40ea4c33a9fbce630842d085363d75aa where I had to fixup a few things for flake8 - - this can be tested locally in the future by running the `make check` target.

Almenon:
Ah, sorry about the .vscode directory. I didn't intend to include that in my PR. Want me to delete the stale branch?

davvid:
No worries.  I don't mind having the stale branch around personally, but feel free to remove it if you prefer.  It actually makes it kinda easier to see where the PRs originated IMO:

image.

PR #285, opened 2020-01-11
speed improvement for long lists of primitives

Almenon:
I was wondering why `x = [[i for i in range(10000000)]]` took so long to calculate in [arepl](https://github.com/Almenon/AREPL-vscode) compared to native python. I used cProfile+snakeviz and found out that jsonpickle was taking up most of the time. 

Before fix:

After fix:

Jsonpickle does some recursion checks to make sure it doesn't recurse forever. By checking if the type is a non-recursive type and flattening immediately if so there is a ~70% performance gain.

Hopefully this change doesn't break anything. I still need to test this a bit more.

### DIFF
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -223,6 +223,24 @@ class Pickler(object):
         return self._flatten(obj)
 
     def _flatten(self, obj):
+
+        #########################################
+        # if obj is nonrecursive return immediately
+        # for performance reasons we don't want to do recursive checks
+        if PY2 and isinstance(obj, types.FileType):
+            return self._flatten_file(obj)
+
+        if util.is_bytes(obj):
+            return self._flatten_bytestring(obj)
+
+        if util.is_primitive(obj):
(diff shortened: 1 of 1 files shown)


Almenon:
Just one test failure in ecdsa_test ..aaggghhh so close

davvid:
This is super cool!  Please rebase -- it looks like the travis environment got updated in the last month and ecdsa went up to v0.15.  I implemented a handler for `threading.Lock` and that got the ecdsa tests passing again.

I would recommend installing tox, python2.7, and any modern python3.6 (e.g. 3.7) and then using the `make tox` target to test them.

Almenon:
Just merged this into arepl backend. Tests are passing there. Note that this might be a breaking change for users who subclass jsonpickle.pickler.Pickler and override _get_flattener

I changed my override of _get_flattener to a override of _flatten:
https://github.com/Almenon/AREPL-backend/pull/101/commits/7e924d62344552d705eb72049592126364e94d14

PR #291, opened 2020-02-23
Declare pep518 build requirements.

### LINKED ISSUE #290: Implicit build dependency on setuptools
This package has an implicit build dependency on setuptools, meaning it will fail to build in an environment where setuptools isn't pre-installed. This issue can be addressed by providing PEP 518 build requirements.

jaraco:
Fixes #290

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -6,3 +6,4 @@ include tests/*.py
 include docs/source/**
 include jsonpickleJS/**
 include *.txt
+include pyproject.toml
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,2 @@
+[build-system]
+requires = ["setuptools>=42"]


jaraco:
Also, if you like this change, I've got many more best-practices that I've been collecting in [jaraco/skeleton](/jaraco/skeleton). See the [the docs there](https://github.com/jaraco/skeleton/blob/master/skeleton.md) for more background. If you're interested in applying those changes wholesale, I can do that (and reconcile the conflicts), after which this project could diverge from those practices. Or if you'd rather cherry-pick different changes, I'd invite that as well, but that would have to come from some other mind that has preferences on which changes to select and omit.

davvid:
Thanks that looks pretty great.  I've had ideas of doing something similar in the past (I think I even called it `skeletor`, before throwing it all away) so skeleton be really great to adopt IMO.

PR #292, opened 2020-02-28
Python 3.8

davvid:
Add support for Python 3.8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ python:
   - &latest_py3 "3.7"
   - "3.8"
 install: pip install tox-travis
-script: make tox
+script: make tox flags='--parallel 0'
 
 jobs:
   fast_finish: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ python:
 # - "3.5" # ecdsa tests fail in 3.5.
   - "3.6"
   - &latest_py3 "3.7"
+  - "3.8"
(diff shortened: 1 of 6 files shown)


hartwork:
Did you forget to add `3.8` to `.travis.yml`?
https://github.com/jsonpickle/jsonpickle/blob/a52ba426c432800ec272f073b357b2998a619666/.travis.yml#L4-L9

davvid:
@hartwork thanks for the heads-up.  I just added it -- wasn't sure if travis supported it yet.

.travis.yml
```diff
@@ -7,8 +7,9 @@ python:
 # - "3.5" # ecdsa tests fail in 3.5.
   - "3.6"
   - &latest_py3 "3.7"
+  - "3.8"
```
hartwork: Are you aware that now `latest_py3` no longer points to the latest Python 3?

davvid: Yeah, I did that on purpose.  python3.8 is so new that I didn't trust `twine` and the deployment steps to work in that environment.  My not-great metric is that py3.8 required code updates for JP itself, and we're not even that big of a library, so I opted to hold that back for stability purposes.

We'll update that eventually, though.  Thanks for the heads-up.

hartwork: I think it would be easy to change the variable name `latest_py3` to something less mis-leading in the mean time. Just an idea.

PR #295, opened 2020-03-03
Make readme warn about security implications (fixes #294)

### LINKED ISSUE #294: Security compared to pickle too hard to find
Hi!

I tried to find an answer if jsonpickle has the same design security issues that pickle has. I found #179 eventually but it took a while and there is nothing like in the master readme, the place where I'd expect security implications as severe as that. Would you be up for adding a warning to some place clearly visible in the readme?

Thanks for your consideration, Sebastian

hartwork:
Fixes #294

(I would love to use a [warning admonition](https://docutils.sourceforge.io/docs/ref/rst/directives.html#admonitions) for this but [GitHub fails to render them properly](https://github.com/github/markup/issues/68).)

CC @davvid

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -67,7 +67,7 @@ enable the numpy extension by registering its handlers::
     >>> jsonpickle_numpy.register_handlers()
 
 Pandas Support
-=============
+==============
 jsonpickle includes a built-in pandas extension.  If would like to encode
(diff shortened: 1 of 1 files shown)


hartwork:
Damn, GitHub doesn't render this like rst2html does, let me fix it for GitHub syntax...

PR #304, opened 2020-04-22
Include LICENSE in MANIFEST.in

synapticarbors:
The project switched from using the `COPYING` file to `LICENSE` at some point and it is not correctly reflected in the `MANIFEST.in`

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,4 @@
-include COPYING README.rst
+include LICENSE README.rst
 include Rakefile
 include *.py
 include jsonpickle/*.py


synapticarbors:
It actually looks like this is unnecessary as the license file is in the source distribution, probably because of the entry in the `setup.cfg`

davvid:
I believe so, checkout https://github.com/davvid/skeleton and its parent https://github.com/jaraco/skeleton which is what my fork is based upon. 

I think `setuptools_scm` does some magic that includes whatever is in git.  I actually don't know the full details, but I'm glad jaraco figured it out for us.

I'll merge this nonetheless, thanks!  Hmm, maybe the better solution is to get rid of that file altogether... hmmm.. do we even need it anymore these days?

davvid:
I think it's not needed, but just in case, this will be merged, and then deleted.

PR #317, opened 2020-07-23
Add additional unit test for complex np objects

ORippler:
Adds unit test case for edge case of complex np objects I found here #189 

Unfortunately I didn't manage to further simplify my input structures (they are `dicts` with 10 entries of `np.arrays` of shape `(4,)` each) or even construct them from scratch, as the unit test would not fail otherwise...


davvid:
Thanks, but the key part is missing here.   There's interesting data in those string keys, but it'd be good to see the code that generated those strings.

You mention a dict with `np.array`s, what's the construction and encoding of that structure look like?


I'd be curious to debug it end-to-end, but all we have is the `decode()` side of the `encode()/decode()` cycle here, it seems.

ORippler:
Hi,

Unfortunately I am currently unable to share the code that generated the strings for `dict_1` and `dict_2`, but I may be able to supply that in ~6 months if everything pans out and my research work is accepted.

However I believe that the MWE provided in my test case is still sufficient. Both `dict_1` and `dict_2` (which are the `dicts` with `np.array`s I refer to initially above) can be decoded based on the string representations provided in the test case.

Furthermore, both `dict_1` and `dict_2` [pass the call to `self.roundtrip`](https://github.com/jsonpickle/jsonpickle/blob/263c8e1f79b7777ed7bf5ae66d8685d16e84bca7/tests/numpy_test.py#L390) on their own after being decoded, which consists of a full `encode()/decode()` cycle. However, when they are joined in e.g. a `tuple` or `list` structure, [they fail with the `_IDProxy` `AttributeError`](https://github.com/jsonpickle/jsonpickle/blob/263c8e1f79b7777ed7bf5ae66d8685d16e84bca7/tests/numpy_test.py#L395) in the `encode()/decode()` cycle.

Therefore, you should be able to fully debug and contrast the behavior of `encode()/decode()` as necessary between the three calls to `self.roundtrip` in the provided test case, which only fails once with the aforementioned error type (unless I am missing something).

Best,

davvid:
There have been a lot of changes recently that improved the handling for dicts inside tuples and lists (which is the use case that seems to be in play here) so I'm inclined to say that this issue is resolved.

Unfortunately I can't add this test case as-is because it embeds the pickled implementation details that will be a maintenance burden to support across jsonpickle and numpy versions, so we won't be merging this for now.  That said, if you could please test the latest in git it would be very helpful to know if your use case has been solved now.

BTW, are you using dictionaries with integer keys?  If so, you may need to specify `keys=True` when encoding and decoding to enable support for that feature.

Thanks for the heads-up!  Hopefully we're all good across the board now thanks to recent development.

Issue #235, opened 2018-10-10
Enum not serializing properly, serializing as string


pmmcmullen94:
I'm using JSONPickle 0.9.6 and when I serialize a custom enum that looks as follows: 
```
class TSDataType(Enum):
    NONE = 0
    STRING = 1
    INTEGER = 2
    DOUBLE = 3
    BOOLEAN = 4
```
it serializes fine sometimes like this: 
```
"DataType": {
     "py/reduce": [
         {
             "py/type": "Enums.TSDataType"
         },
         {
             "py/tuple": [
                 2
             ]
         },
         null,
         null,
         null
     ]
 },
```
but then other times it serializes like this:
`"DataType": "<TSDataType.INTEGER: 2>",`

The bottom way is serializing as a string whereas the top can successfully be deserialized back to the object. Is there anything that might be causing this that is known? 

My serialization code looks like this:
```
jsonpickle.encode(value, make_refs=False)
```

Thanks in advance

davvid:
Please retest using a modern version.  Enums are well handled these days.  Feel free to reopen if you have a test case that demonstrates an issue that is not currently handled by the current test suite.

numblr:
I have the same issue in version 1.4.1 when make_refs is set to False.

davvid:
Indeed, thank you @pmmcmullen94 and @numblr this was indeed an issue under `make_refs=False`.

I adjusted the behavior so that enums are treated like atoms and are thus never candidates for cycle detection.  A new test case was added to guarantee that this use case works as expected under `make_refs=False`.

This is was fixed in 99af0c8b96b5728ab6cfb4bc44d1d2ca4d9566bf -- please gives it a try and let me know if it improves your use cases.

The fix will be in the upcoming `v1.4.3` release.  I'll tentatively plan to tag the next version before the end of the year.  This change alone warrants a new release IMO, but since we just tagged a new `v1.4.2` just before this commit then it wouldn't hurt to wait a couple of weeks before tagging `v1.4.3`.

PR #327, opened 2020-11-18
Use set for checking types

laike9m:
Replaced tuple with set to make the `in` operation faster.

### DIFF
--- a/jsonpickle/util.py
+++ b/jsonpickle/util.py
@@ -33,7 +33,8 @@ if PY2:
     import __builtin__
 
 SEQUENCES = (list, set, tuple)
-PRIMITIVES = (compat.ustr, bool, type(None)) + numeric_types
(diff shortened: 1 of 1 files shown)


davvid:
Thanks!

laike9m:
Is it possible to release a new version?

davvid:
`jsonpickle==1.4.2` should be available via pypi now.  cheers

laike9m:
Nice, thanks for the work!

davvid:
I should also mention that a fix for #235 was just committed.  Please not that this change was *not* included in 1.4.2 since it was hot off the presses.  I plan to tag a release with that fix before the end of the year, but please shout out if it would be helpful sooner rather than later.

cheers!

laike9m:
Got it, no worries

PR #330, opened 2020-12-14
Enable decode of timedelta pandas dtype

jdow7:


davvid:
Thank you so much!  This looks correct, but it would be really nice if we could have a unit test to make sure that it works correctly, and to make sure that we don't break this functionality in the future.

Sorry for the delay in responding to the PRs -- I have some more time now.  BTW, there is a change brewing in #338 that would be great to test in tandem with these changes.  Could you maybe test that branch with these change to verify that they're all good?

Maybe we can come up with the test case together, in case you're having trouble running the tests.  The `tests/pandas_test.py` is probably the best place to add the new tests.

Let's try to get this tested and then we'll merge this right in.  Thanks again!

davvid:
Ah, sorry I missed your note in #331 -- I can try to rearrange recipe you provided there into a test case as well since you mentioned you're new.  Let me know if you were able to run the test suite, that's usually the first step to writing new tests since it's easier to do once it's up and running.

If I'm able to write a test that fails without these changes and then passes with them present then I'll merge this shortly thereafter.  Thanks again.

davvid:
I added a test case to `tests/pandas_test.py` (near the bottom) with your recipe from #331.  It worked great, thanks again!

PR #339, opened 2021-01-14
Micro-Optimization

Theelgirl:
Currently, if you are trying to decode something with the object tag, it takes 10 ``if`` checks for previous tags. Because three of those are only for backwards compatibility, and most people probably don't use them, I decided to push them closer to the end. This shaves off about 20% of the time per call of ``Unpickler._restore`` on my VPS with some of my test data (mixed 'py/object' and 'py/state'), and it didn't fail any pytest tests.

I plan to submit more optimizations in the future, I see great potential in this library for use if speed and security are taken care of.

### DIFF
--- a/jsonpickle/unpickler.py
+++ b/jsonpickle/unpickler.py
@@ -15,7 +15,6 @@ from . import handlers
 from .compat import numeric_types
 from .backend import json
 
-
 def decode(
     string, backend=None, context=None, keys=False, reset=True, safe=False, classes=None
 ):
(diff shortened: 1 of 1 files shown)


davvid:
This sound really cool!  Let me know if you need any help getting the test suite running locally for verification.  Hopefully there's not some use cases that are reliant on the current order of tag checking or other things.  Thanks for your help.

Theelgirl:
Thanks! I got the testing suite running locally, but I couldn't get flake8 and black running so I could use some advice with that.

Theelgirl:
Also, is it okay if I add some commits for documentation updates to this? I noticed that docs for Unpickler are very sparse in terms of describing what the various decoding options do, and I'm not sure how to make a new PR with the documentation changes but without these optimization changes.

Theelgirl:
Closing in favor of an even more efficient unpickler that I'm going to make a new PR for. I tested this one with tox on every python version, so I expect it to work.

PR #340, opened 2021-01-16
Large optimization to unpickler

Theelgirl:
I realized that in certain cases, actually a sizeable number of cases, we don't need to check if it has a tag because it's a float, or an int, or something else that can't contain a string tag. This doubles the speed of unpickling for my personal use case (which involves a large amount of ints/floats), and should provide a statistically significant speed increase for most other cases also.

### DIFF
--- a/jsonpickle/unpickler.py
+++ b/jsonpickle/unpickler.py
@@ -176,28 +176,49 @@ class Unpickler(object):
         self._proxies = []
 
     def _restore(self, obj):
+        # if obj isn't in these types, neither it nor nothing in it can have a tag
+        # don't change the tuple of types to a set, it won't work with isinstance
+        if not isinstance(obj, (str, list, dict, set, tuple)):
+
+            def restore(x):
+                return x
+
+        else:
+            restore = self._restore_tags(obj)
+        return restore(obj)
+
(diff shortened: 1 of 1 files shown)


davvid:
Sweet, thanks!

Theelgirl:
No problem! I was looking over encoder optimizations recently, do you think I should focus more on optimizing encoding or decoding?

davvid:
Hmm that's a great question.  I'm not really sure which has the most low-hanging fruit.  `decode()` performance might benefit more users would be my guess.  For example, an app that stores state may read it more often then it stores it, but like most things it's fairly app specific.

There's a bunch of backwards-compatibility code around the `self._namestack`, `self._namedict` and `self._refname()` in support of the older `tags.REF` stuff that was emitted by much older versions.  There's already enough stuff since `v1.5.0` that we can tag `v1.5.1` shortly to clear the way for another minor (`v1.6.0`) that drops that backwards-compatibility support for `tags.REF` if it improves performance even further.

Theelgirl:
Just profiled that backwards-compatibility code, it doesn't seem to take up much cpu time or memory so I think I'll leave it for now.

PR #343, opened 2021-01-17
4x speedup for is_reducible

Theelgirl:
Speeds up encode by about 30% on my machine/test data, plus passes all the tests on my machine.


Theelgirl:
Clarification: It sped up encode by a relatively large percentage because my test data was encoding a custom class with ``__getstate__``, ``__setstate__``, and ``_slots__`` defined, when I run it on a version without those defined, it's less of a speedup (though I haven't fully determined how much less).

PR #345, opened 2021-01-18
Small speedups for _flatten_impl and others

Theelgirl:
Speeds up ``Pickler._flatten_impl``, ``Pickler._get_flattener``, and ``util._is_function``.

``_flatten_impl``:
- Condense type checks
- Remove function call

``_get_flattener``:
- Removed slow function alias
- Rearranged checks to check more common cases first

``is_function``:
- Avoided construction of a new set every call

### DIFF
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -271,15 +271,14 @@ class Pickler(object):
         if PY2 and isinstance(obj, types.FileType):
             return self._flatten_file(obj)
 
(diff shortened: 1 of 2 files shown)


Theelgirl:
Hmm, it looks like this breaks 2.7 so it'd be better for v2.0

Theelgirl:
@davvid Thoughts? I fixed it for 2.7

Issue #346, opened 2021-01-22
Fails for multilevel columns pandas DataFrames


antoinecollet5:
Jsonpickle does not support multilevel columns pandas DataFrames.

Here is an example that fails. As you can see, the multiindex is supported, not the multilevel columns.
``` python
import pandas as pd
import numpy as np
import jsonpickle as jp
import jsonpickle.ext.pandas as jsonpickle_pd

### This works ###
midx = pd.MultiIndex(levels=[["zero", "one"], ["x", "y"]], 
                     codes=[[1, 1, 0, 0], [1, 0, 1, 0]])
df = pd.DataFrame(np.random.randn(4, 2), index=midx, columns=['col1', 'col2'])
frozen = jp.encode(df)
thawed = jp.decode(frozen)

### This does not ###
iterables = [['inj', 'prod'], ['hourly', 'cum']]
names = ['first', 'second']
# transform it to tuples
columns = pd.MultiIndex.from_product(iterables, names=names)
# build a multi-index from it
df2 = pd.DataFrame(np.random.randn(3, 4), index=["A", "B", "C"], 
                   columns=columns)
frozen2 = jp.encode(df2)
thawed2 = jp.decode(frozen2)
```

Here is a modification of pandas.py that seems to fix the issue. I am new to git so I need to figure out how to make a clean pull request. I ll try to add an appropriate unit test.

``` python
def make_read_csv_params(meta):
    meta_dtypes = meta.get('dtypes', {})
    # [None] makes it compatible with objects serialized before 
    # column_levels_names has been introduced.
    column_level_names = meta.get('column_level_names', [None])
    # The header is used to select the rows of the csv from which
    # the columns names are retrived
    header = meta.get('header', [0])
    parse_dates = []
    converters = {}
    dtype = {}
    timedeltas = []
    for k, v in meta_dtypes.items():
        if v.startswith('datetime'):
            parse_dates.append(k)
        elif v.startswith('complex'):
            converters[k] = complex
        elif v.startswith('timedelta'):
            timedeltas.append(k)
            dtype[k] = 'object'
        else:
            dtype[k] = v

    return dict(dtype=dtype, header=header, parse_dates=parse_dates, 
                converters=converters), timedeltas, column_level_names


class PandasDfHandler(BaseHandler):
    pp = PandasProcessor()

    def flatten(self, obj, data):
        dtype = obj.dtypes.to_dict()

[...]

Issue #336, opened 2021-01-06
Fails to built with numpy 1.20.0, py 3.10 or 3.9.


limburgher:
https://bugzilla.redhat.com/show_bug.cgi?id=1913439

python-jsonpickle fails to build with Python 3.10.0a4 and Python 3.9.1. This seems related to the updated NumPy.

______________________ NumpyTestCase.test_dtype_roundtrip ______________________

self = <numpy_test.NumpyTestCase testMethod=test_dtype_roundtrip>

    def test_dtype_roundtrip(self):
        if self.should_skip:
            return self.skip('numpy is not importable')
        dtypes = [
            np.int,
            np.float,
            np.complex,
            np.int32,
            np.str,
            np.object,
            np.unicode,
            np.dtype('f4,i4,f2,i1'),
            np.dtype(('f4', 'i4'), ('f2', 'i1')),
            np.dtype('1i4', align=True),
            np.dtype('M8[7D]'),
            np.dtype(
                {
                    'names': ['f0', 'f1', 'f2'],
                    'formats': ['<u4', '<u2', '<u2'],
                    'offsets': [0, 0, 2],
                },
                align=True,
            ),
        ]
    
        if not PY2:
            dtypes.extend(
                [
                    np.dtype([('f0', 'i4'), ('f2', 'i1')]),
                    np.dtype(
                        [
                            (
                                'top',
                                [
                                    ('tiles', ('>f4', (64, 64)), (1,)),
                                    ('rtile', '>f4', (64, 36)),
                                ],
                                (3,),
                            ),
                            (
                                'bottom',
                                [
                                    ('bleft', ('>f4', (8, 64)), (1,)),
                                    ('bright', '>f4', (8, 36)),
                                ],
                            ),
                        ]
                    ),
                ]
            )
    
        for dtype in dtypes:
>           self.assertEqual(self.roundtrip(dtype), dtype)

tests/numpy_test.py:95: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib64/python3.10/site-packages/numpy/core/_internal.py:61: in _usefields
    names, formats, offsets, titles = _makenames_list(adict, align)
[...]

PR #347, opened 2021-01-22
Update pandas.py

antoinecollet5:
Makes it compatible with multi-level columns Pandas DataFrame.
The trick is to use the keyword `header` from `pandas.read_csv` to indicate how many rows of the `csv` file to take into account for the columns creation.
The names of the levels are also stored and reassigned after the df has been created.

### DIFF
--- a/jsonpickle/ext/pandas.py
+++ b/jsonpickle/ext/pandas.py
@@ -55,7 +55,12 @@ class PandasProcessor(object):
 
 def make_read_csv_params(meta):
     meta_dtypes = meta.get('dtypes', {})
-
+    # [None] makes it compatible with objects serialized before 
+    # column_levels_names has been introduced.
+    column_level_names = meta.get('column_level_names', [None])
+    # The header is used to select the rows of the csv from which
+    # the columns names are retrived
+    header = meta.get('header', [0])
     parse_dates = []
     converters = {}
     dtype = {}
@@ -71,7 +76,8 @@ def make_read_csv_params(meta):
         else:
             dtype[k] = v
 
-    return dict(dtype=dtype, parse_dates=parse_dates, converters=converters), timedeltas
+    return dict(dtype=dtype, header=header, parse_dates=parse_dates, 
+                converters=converters), timedeltas, column_level_names
 
 
 class PandasDfHandler(BaseHandler):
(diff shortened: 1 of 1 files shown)


Theelgirl:
Just want to make it clear for any maintainers coming along that according to travis, the tests that this fails are only flake8 and black tests, not compatibility tests, so the "errors" could be fixed with a simple flake8 and black runthrough.

davvid:
Thank you @Theelgirl and @antoinecollet5  -- I'll tag the repo now and start merging this down as the next round of topics for the next release.  Since these are "just" formatting issues I'll fix that up locally.  Cheers!

davvid:
There is one test failure where this change dropped support for timedelta columns.  I'm restoring that functionality now; it's a simple fix.

I'll also be adding a unit test that exercises the issue described in #346 so that we don't regress on this enhancement in the future.

davvid:
I merged this in after adding a unit test that verifies the behavior.

There are a few remaining issues.  It looks like this functionality requires that the `ujson` module is available.  I hadn't noticed because the virtualenvs I use for testing all had it available.  I tweaked the requirements-dev.txt to reflect that fact that ujson is available for all python versions now.

This also led me to realize that, indeed, the tests do fail when numpy 1.20.0 is installed, as reported in #336.

The setup.cfg file is currently locked down to numpy 1.19 for now, but this would be great to improve.  Likewise, it would be a good idea to make this new feature not require ujson.

The root cause seems to be that this feature relies on sticking tuples in dictionary keys.  While we do support that in jsonpickle, the flatten process shouldn't leave that around since it will cause issues since it's non-standard json.  It's the dtypes dict that ends up containing tuples which is what makes it go sideways.  I'll be looking into resolving this one.

The numpy ujson settings in `setup.cfg` will have to be tweaked (and your `.tox` directory purged) in order to see the numpy 1.20 failures.. it's currently locked down to 1.19.4.

davvid:
Alrighty, I was able to eliminate the ujson oversight by ensuring that the `dtypes` value in the meta dict is now flattened before handling off to the backend.  Please see 39a6cbed3b9d99b9d3ce560eed44a2c10e1e9136 for the small tweak that made that possible.

Thanks again!

antoinecollet5:
Thank you @davvid ! Cheers

PR #348, opened 2021-01-29
Update python versions for tox targets

Theelgirl:
Add python 3.9 as a target since it's now the current stable release, and increment the lib testing by one python version since we incremented the main testing versions by 1.


davvid:
I should have merged this earlier before the numpy stuff but it's all good.  Thanks again!

PR #352, opened 2021-02-08
Create basic benchmarks

Theelgirl:
I will update BENCHMARKS.md as I run more.


Theelgirl:
Wait don't merge yet: the jsonpickle_benchmarks script at the moment can only be run outside of the jsonpickle home directory, as pytext.ini interferes with it by adding options that it doesn't recognize. Anyone is welcome to fix it, [here are the docs.](https://pytest-benchmark.readthedocs.io/en/latest/usage.html)

Theelgirl:
@davvid How does this look?

PR #353, opened 2021-02-11
Speed up is_reducible by about 30%

Theelgirl:
Also add a note explaining how to achieve further speedups, but that would require disabling black's line length checker.

### DIFF
--- a/jsonpickle/util.py
+++ b/jsonpickle/util.py
@@ -381,18 +381,13 @@ def is_reducible(obj):
     # defaultdicts may contain functions which we cannot serialise
     if is_collections(obj) and not isinstance(obj, collections.defaultdict):
         return True
-    # sets are slightly slower in this case
-    if type(obj) in NON_REDUCIBLE_TYPES:
+    # combining all this into a single long if statement with many ORs
+    # would be significantly faster but it'd be over the 88 char line limit :(
+    if type(obj) in NON_REDUCIBLE_TYPES or obj is object:
         return False
-    elif obj is object:
+    elif is_dictionary_subclass(obj) or isinstance(obj, types.ModuleType):
         return False
-    elif is_list_like(obj):
-        return False
-    elif isinstance(obj, types.ModuleType):
-        return False
-    elif is_dictionary_subclass(obj):
-        return False
-    elif is_reducible_sequence_subclass(obj):
+    elif is_reducible_sequence_subclass(obj) or is_list_like(obj):
         return False
     elif isinstance(getattr(obj, '__slots__', None), iterator_types):
         return False


Theelgirl:
On that note, we could massively speed up a bunch of util functions if we disable Black for that file, as putting expressions onto a single line seems to make the job much easier and quicker for the Python parser. What do you think davvid?

jsonpickle/util.py
```diff
@@ -381,18 +381,13 @@ def is_reducible(obj):
     # defaultdicts may contain functions which we cannot serialise
     if is_collections(obj) and not isinstance(obj, collections.defaultdict):
         return True
-    # sets are slightly slower in this case
-    if type(obj) in NON_REDUCIBLE_TYPES:
+    # combining all this into a single long if statement with many ORs
+    # would be significantly faster but it'd be over the 88 char line limit :(
+    if type(obj) in NON_REDUCIBLE_TYPES or obj is object:
```
davvid, line 386: Ah that's an interesting note about combining the conditionals.  There are approaches we can take to improve that without having to impact the formatting.

If we did need to keep a specific format, there are `# fmt: on` and `# fmt: off` directives we can use: https://github.com/psf/black#the-black-code-style

There are also `pylint: disable=line-too-long` for pylint and `# noqa: E501` for flake8 to disable length checks.

That said, it's usually best if we could do this while still staying within the limits.  One way is that Python does allow the conditionals to run over multiple lines, for example, instead of:

```
if foo and bar and baz:
```

We can write:
```
if (foo
    and bar
    and baz
):
```

and Python's parser will allow it because we've put the conditional expression inside `(parentheses)` which allows the expression to extend over multiple lines.

We could even do:

```
if foo \
    and bar \
    and baz:
```

but we try to avoid backslashes because they're visually noisy.  Another reason we avoid backslashes is because if someone accidentally adds a trailing space (which is pretty hard to spot, since it's a space) after the backslash it breaks the expression.  Thus, parentheses are usually the go-to for this kind of situation.

I'd be curious to see if breaking it up into multiple lines affects performance.  I wouldn't expect it to, but it is Python so I wouldn't be surprised either.

Theelgirl: Actually, I tested breaking it into multiple lines. That actually results in a net increase in time taken over the original version. I believe this is reflecting the time python takes to parse the lines, but I'm unsure.

Theelgirl:
```py
   385     15020      58703.0      3.9     63.4      if type(obj) in NON_REDUCIBLE_TYPES or obj is object or is_list_like(obj) or isinstance(obj, types.ModuleType) or is_dictionary_subclass(obj) or is_reducible_sequence_subclass(obj) or isinstance(getattr(obj, '__slots__', None), iterator_types) or (is_type(obj) and obj.__module__ == 'datetime'):
   386      5004       2733.0      0.5      3.0          return False
   387     10016       6157.0      0.6      6.7      return True
```

```py
   385     95185      56596.0      0.6     28.5      if (
   386     15025      12479.0      0.8      6.3          type(obj) in NON_REDUCIBLE_TYPES
   387     15025       9373.0      0.6      4.7          or obj is object
   388     10020      18136.0      1.8      9.1          or is_list_like(obj)
   389     10020       8875.0      0.9      4.5          or isinstance(obj, types.ModuleType)
   390     10020      18204.0      1.8      9.2          or is_dictionary_subclass(obj)
   391     10020      13662.0      1.4      6.9          or is_reducible_sequence_subclass(obj)
   392     10020      12262.0      1.2      6.2          or isinstance(getattr(obj, '__slots__', None), iterator_types)
   393     15035      15029.0      1.0      7.6          or (is_type(obj) and obj.__module__ == 'datetime')
   394                                               ):
   395      5005       2894.0      0.6      1.5          return False
   396     10020       5618.0      0.6      2.8      return True
```

line | number of times executed | total time taken | time per execution | percent time taken

davvid:
Ah, I see, thanks for trying that out.  Heh, I guess Python's bytecode representation isn't helping much for this case. Since breaking it apart didn't pan out then using the `# fmt: off/on` directive and comments to disable pylint/flake8 would be fine since it results in further enhancements.

Thanks again!

PR #360, opened 2021-06-24
Replace sklearn with scikit-learn

hugovk:
The sklearn package is a wrapper around scikit-learn, and sklearn will be removed at some point, starting with brownouts.

Instead, use scikit-learn directly.

See https://github.com/scikit-learn/scikit-learn/issues/8215 for more info.


hugovk:
See https://github.com/jsonpickle/jsonpickle/pull/361 to fix the CI.

PR #361, opened 2021-06-24
Replace Travis CI with GitHub Actions

hugovk:
travis-ci.org has finally gone offline 6 months after the originally planned end date.

image

Either the build needs migrating to travis-ci.com, or to another CI like GitHub Actions.

GHA has 20 parallel jobs compared to 5 on Travis, and I've found it very flexible and an improvement over Travis CI, without mentioning the new Travis build quota problems.

See https://github.com/hugovk/jsonpickle/actions/runs/967938334 for an example build.


Theelgirl:
I'd like to second this, switching from Travis CI is very important. I was thinking about working on a bitflag implementation of jsonpickle that I might eventually PR here, but it needs to be testable by something.

davvid:
👍 thank you so much for migrating us over to github actions.

hugovk:
You're welcome, thank you for maintaining jsonpickle!

PR #372, opened 2021-12-23
Add ``v1_decode`` argument to ``jsonpickle.decode``

### LINKED ISSUE #364: jsonpickle 2.0.0 fails to decode pickled files created by jsonpickle 1.4.1
I've been attempting to upgrade the jsonpickle package in our project, where we store a number of files in pickled format. It happens that in some UTs for that area their evaluation failed, when I only change the jsonpickle package to 2.0.0.  For 1.5.2 it still works. I also locally tested with the real data, in case it would be case of overly specific UTs. But those ended up with corrupted JSON data. As one can see in the screenshotRampdown-pickle-issue, instead of 

```
"previousFeatureTeam": {
        "py/id": 1
      },
```

on the left side created with 1.5.2, the version created by jsonpickle suddenly creates a list and effectively doubles the output by reusing existing data (but even that seems corrupted).

I suspect that the format changed or that maybe 2.0.0 introduced a regression, but I couldn't find any information on what actually changed for 2.0.0. Do I need to do something special for conversions from older data?

Theelx:
This allows users to deserialize objects that were encoded with version 1, in version 2. Closes #364.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,8 @@ v2.0.1
       and/or repr objects. (+354)
     * ``is_iterator()`` was sped up by ~20% by removing an unnecessary
       variable assignment. (+354)
+    * ``jsonpickle.decode`` has a new option, ``v1_decode`` to assist in
+      decoding objects created in jsonpickle version 1.(#364)
 
 v2.0.0
 ======
--- a/jsonpickle/unpickler.py
+++ b/jsonpickle/unpickler.py
@@ -17,7 +17,14 @@ from .backend import json
 
 
 def decode(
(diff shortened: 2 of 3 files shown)


Theelx:
I'm ignoring the test failure for demjson on 3.9, as demjson hasn't been updated since 2015 and there's an open PR on their github repo saying it doesn't have compatibility with 3.9+ (https://github.com/dmeranda/demjson/pull/34). It might be worth looking into a jsonpickle 3.0.0 dropping support for all the outdated parts of the codebase (python 2, and <=3.6, as 3.6 is about to hit EOL).

PR #373, opened 2022-01-07
Update docs for unpickleable argument to encode

### LINKED ISSUE #366: incorrect serialization of complex object containing sklearn instances
Hello there.

During our usage of jsonpickle library in combination with sklearn model training and (de)serialization we came across an object, which jsonpickle cannot correctly serialiaze and deserialize. More specifically: An object in memory shows correct behavior while in memory, but it's ill-built and its behavior is buggy after serialization and deserialization.

Looking through open issues this seems to be related to issue #181
The only difference to our case is that we work with the default setting `make_ref=False`.

Our code:

```python
import jsonpickle
import jsonpickle.ext.numpy as jsonpickle_np
import jsonpickle.ext.pandas as jsonpickle_pd
from typing import Text, Any
from sklearn.pipeline import Pipeline


def json_unpickle(file_name: Text) -> Any:
    jsonpickle_np.register_handlers()
    jsonpickle_pd.register_handlers()
    with open(file_name, "r", encoding="utf-8") as f:
        return jsonpickle.loads(f.read())


def json_pickle(file_name: Text, obj: Any) -> None:
    jsonpickle_np.register_handlers()
    jsonpickle_pd.register_handlers()
    with open(file_name, "w", encoding="utf-8") as f:
        f.write(jsonpickle.dumps(obj, indent=1))


class WrappedModel:
    def __init__(self, pipeline: Pipeline):
        self.pipeline = pipeline
        self.classes_ = self.pipeline.classes_


# load sklearn pipeline components
vectorizer = json_unpickle("vectorizer.json")
sgd = json_unpickle("sgd.json")
# build a WrappedModel instance
sklearn_pipeline = Pipeline(steps=[vectorizer, sgd])
wrapped_model = WrappedModel(pipeline=sklearn_pipeline)
print("output:")
print(wrapped_model.classes_)
# pickle & unpickle
json_pickle("model.json", wrapped_model)
wm = json_unpickle("model.json")
print("output after serialization + deserialization:")
print(wm.classes_)
```

Outputs:
```
$ python ./run.py
output:
['i_goodbye' 'i_greet' 'i_how_are_you' 'i_my_name_is' 'i_nice_day' 'i_no'
 'i_restart' 'i_select_animal' 'i_swearing' 'i_thanks' 'i_unsupported'
 'i_who_are_you' 'i_yes']
output after serialization + deserialization:
[[-4.6257104 ]
 [-3.24113826]
 [-3.22166714]
 [-3.57920356]
 [-2.45528723]
 [-2.56593293]
 [-3.75261252]
 [-2.61424524]
 [ 9.18493755]
 [-4.17989477]
 [-4.02656936]
 [-4.08638535]
 [-4.06015507]]


$ pip list
Package         Version
--------------- -------
joblib          1.1.0
jsonpickle      2.0.0
numpy           1.21.3
pandas          1.3.4
pip             21.3.1
python-dateutil 2.8.2
pytz            2021.3
scikit-learn    0.24.2
scipy           1.7.1
setuptools      58.3.0
six             1.16.0
threadpoolctl   3.0.0
wheel           0.37.0

$ python
Python 3.8.12 (default, Sep 10 2021, 00:16:05)
[GCC 7.5.0] on linux

$ uname -a
Linux abc 5.4.0-89-generic #100~18.04.1-Ubuntu SMP Wed Sep 29 10:59:42 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

This behavior was observed on an ubuntu linux machine (all above outputs), but we were able to reproduce the same buggy behavior in a mac os environment: python 3.8, with the same python packages installed as mentioned above.

**Important:** Please download [files.zip](https://github.com/jsonpickle/jsonpickle/files/7409094/files.zip) mentioned in the code to reproduce the bug.

Random, but relevant notes:
- It was very difficult to construct a sensible minimal example. Unfortunatelly, we were not able to get rid of the two .json files: `vectorizer.json` + `sgd.json`
- `vectorizer.json` + `sgd.json` do not contain binary-encoded stuff, so they are safe to import in other environments to reproduce the buggy behavior.
- We investigated the objects in `vectorizer.json` + `sgd.json` and we observed only correct behavior with them, i.e., their (de)serialization does not change them and we obtain the same object by construction "in memory", "the ordinary python way" as by deserializing them from files `vectorizer.json` + `sgd.json`.
- Indeed, it seems that the `"py/id": 37` in the output `model.json` is not computed correctly. We were able to "point it" to the correct object (= the array of strings) by setting `"py/id": 35`. Interestingly, this suggests that the index is "off by 2" instead of "off by 1" as mentioned in issue #181
- We observed the same buggy behavior with version `jsonpickle ==1.5.2`.

Theelx:
Closes #366.

### DIFF
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -43,10 +43,10 @@ def encode(
         information necessary to turn the JSON data back into Python objects,
         but a simpler JSON stream is produced. It's recommended to set this
         parameter to ``False`` when your code does not rely on two objects
-        having the same ``id()`` value and those two objects being equal using
-        ``==`` is sufficient, such as when serializing sklearn instances. If you
-        experience (de)serialization being incorrect when you use numpy, pandas,
-        or sklearn handlers, this should be set to ``False``.
+        having the same ``id()`` value, and when it is sufficient for those two
+        objects to be equal by ``==``, such as when serializing sklearn
+        instances. If you experience (de)serialization being incorrect when you
+        use numpy, pandas, or sklearn handlers, this should be set to ``False``.
     :param max_depth: If set to a non-negative integer then jsonpickle will
         not recurse deeper than 'max_depth' steps into the object.  Anything
         deeper than 'max_depth' is represented using a Python repr() of the
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -39,9 +39,14 @@ def encode(
 ):
     """Return a JSON formatted representation of value, a Python object.
 
-    :param unpicklable: If set to False then the output will not contain the
+    :param unpicklable: If set to ``False`` then the output will not contain the
         information necessary to turn the JSON data back into Python objects,
-        but a simpler JSON stream is produced.
+        but a simpler JSON stream is produced. It's recommended to set this
+        parameter to ``False`` when your code does not rely on two objects
+        having the same ``id()`` value and those two objects being equal using
+        ``==`` is sufficient, such as when serializing sklearn instances. If you
+        experience (de)serialization being incorrect when you use numpy, pandas,
+        or sklearn handlers, this should be set to ``False``.
     :param max_depth: If set to a non-negative integer then jsonpickle will
         not recurse deeper than 'max_depth' steps into the object.  Anything
         deeper than 'max_depth' is represented using a Python repr() of the


Theelx:
Ignoring demjson-related test failures on 3.9, as it doesn't officially support 3.9.

PR #374, opened 2022-01-07
Remove DemJSON Test

Theelx:
DemJSON doesn't support CPython 3.9+. It'd be worth considering dropping its backend, as DemJSON itself hasn't been updated since 2015.

### DIFF
--- a/tests/backend_test.py
+++ b/tests/backend_test.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, division, unicode_literals
 import decimal
+import sys
 import unittest
 from warnings import warn
 
@@ -30,6 +31,8 @@ class BackendBase(SkippableTest):
 
         self._is_installed(backend)
 
+        self.backend = backend
+
         jsonpickle.load_backend(*args)
         jsonpickle.set_preferred_backend(backend)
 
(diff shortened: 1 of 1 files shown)


Theelx:
@davvid I'm in the process of fixing up all the tests and removing obsolete code. Is it okay to remove the DemJSON backend considering its last update date? Next, I plan to work on fixing the FutureWarning for the pandas tests, and fixing some DeprecationWarnings for the numpy dtype tests.

Theelx:
I'm going to merge this, and hopefully fix the FutureWarning and DeprecationWarnings sometime this week.

davvid:
Users that were using demjson may have been relying on its non-standard handling of comments. It's made me wonder whether we should support yaml backends instead. (not related to this cleanup, just thinking out loud)

In a sense, the core problem in jsonpickle could almost be thought of as a dictpickle, and we can support multiple ways to de/serialize from/to that dict.

If users really need comments, relying on a yaml backend sounds like a better long-term solution than a non-standard json parser. Of course, if demjson ever starts getting updated then we can always revisit this decision.

Thanks for cleaning out the demjson backend.

PR #376, opened 2022-01-09
Add 3.10 support

Theelx:
This will:
- Add CPython 3.10 to the test suite
- Clarify on PyPi which versions of CPython 3 are supported
- Remove testing of SQLAlchemy >=1.0,<1.2 in order to make tests pass with CPython 3.10

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["2.7", "3.6", "3.7", "3.8", "3.9"]
+        python-version: ["2.7", "3.6", "3.7", "3.8", "3.9", "3.10"]
         os: [ubuntu-20.04]
         env:
           - TOXENV: libs
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,8 +15,6 @@ jobs:
         os: [ubuntu-20.04]
         env:
           - TOXENV: libs
-          - TOXENV: sa10
-          - TOXENV: sa11
           - TOXENV: sa12
           - TOXENV: sa13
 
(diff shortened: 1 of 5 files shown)


davvid (approved):

davvid:
Now that this is merged, would this be a minor bump according to semver? I'll update the changelog now.

davvid:
I updated `CHANGES.rst` with a message that's ready to get tagged as v2.1.0.

Let me know if we should also tag an earlier commit in this history as possibly v2.0.1 or not.

b007bc53dd9e097c829ac503a4a728fec3c11825, for example, could potentially be tagged `v2.0.1` and then I can separate out the later changes as being v2.1.0-only; but maybe it's not worth it?

I think technically the `v1_decode` flag is a new feature, hence the minor bump, but if we think it's okay adding that to a patch version I can tweak that back to be `v2.0.1` instead before tagging it. It seems like most semver guides suggest that this should be minor.

Let me know what you think. If we're cool with a minor bump then I'll go ahead and tag `main` as-is.

Theelx:
Yeah, I think this is ready for 2.1.0. I believe I had done some optimization work between 2.0 and when I checked back on here in mid-December, so if there was a 2.0.1, then those optimizations would be included in a bugfix release. I think that'd be acceptable according to SemVer, but I still think optimizations would be better as a minor release rather than bugfix.

davvid:
also, if you don't mind the shoutout, kindly add a link to your profile to `docs/index.rst` so that you show up in the Authors section ;-)

Theelx:
Ah, thanks for letting me do that! I accidentally pressed enter when I was deciding whether or not to include my email, so it automatically pushed it. I'm going to scrub out my email from the commit history locally, and then overwrite it. I'd appreciate if you could wait to do any more commits until I do that.

Theelx:
Ok done, I seem to have added two more commits than necessary (I haven't used BFG repo cleaner in a long time). Thanks for waiting!

PR #384, opened 2022-05-05
Add on_missing argument to Unpickler

### LINKED ISSUE #190: Throw exception when target class is not defined
When running `jsonpickle.decode()` and the target class denoted by `py/object` does not exist, a silent failure will occur. Quick demo:

```python
import jsonpickle

class Test:
    def __init__(self, i):
        self.i = i

frozen = jsonpickle.encode(Test(4))
print(jsonpickle.decode(frozen))
del Test
print(jsonpickle.decode(frozen))
```
Will output:
```
$ python3 demo.py
<__main__.Test object at 0x7f26fc6af780>
{'py/object': '__main__.Test', 'i': 4}
```
This makes troubleshooting hard when trying to access members of the class later. Normally something like this would occur when saving a file while running a module as a script and later trying to load that file while running a different script that only imports this module.

I propose an additional keyword argument to `decode()`, maybe `strict=False`. If this was set to true, an Exception should be thrown when [_restore_object()](https://github.com/jsonpickle/jsonpickle/blob/6a062399190f1f978b81f68471d2f284f80834f0/jsonpickle/unpickler.py#L143) can't find the specified class. [`NameError`](https://docs.python.org/3/library/exceptions.html#NameError) seems like a suitable one.

### LINKED ISSUE #313: jsonpickle returns a <dict> object without error or warning when a class is not instantiable because a module it uses is not found.
Greetings! 

I use jsonpickle to persist a class to be transported for a Spark job.  In that environment, I have tried to keep in parity with modules used vs modules needing to be transported and added to the PYTHONPATH vs modules available within the spark environment on the executors/workers.

It works! Nice! But ...

As I was testing this flow - I found that one of my test cases was failing with a strange error:

The pickled class' JSON looks like this:
```
{'py/object': 'smoke.test.PersistSampleToLake', '_extra_kwargs': {'rows': '10'}}
```

And when attempting to read the jsonpickle -- it didn't complain or error out:
```
2020-06-22,17:31:08.195 INFO {spark_driver} [<module>] Spark-Driver read pickled class <class 'dict'>
```

However it wasn't the class I thought it was:
```
2020-06-22,17:31:08.195 INFO {spark_driver} [<module>] Executing dict.execute_in_spark()
  File "spark_driver.py", line 32, in <module>
    runner.execute_in_spark()
AttributeError: 'dict' object has no attribute 'execute_in_spark'
``` 
Note that the class it loaded was returned as a ```<dict>``` - not my class "smoke.test.PersistSampleToLake" which DOES have an "execute_in_spark()" function.

In triaging this - I found that there was a module "sklearn" which was being utilized by the author in this case, that wasn't present on the executor / worker machines.  Seems fair that it can't decode a class that is referencing something the pickler can't see.  Quite fair.

What I expected:
XXXXXError - attempting to decode a class using a module we can't find "sklearn"

What I received:
No Warning
No Error
a <dict> object containing my pickle.

Can we enhance the error reporting portion of this tool to be a little more vocal about actual ERRORS which will stop the parsing and just hand you a dict?

I spent hours trying to find this one - where a single 1 line logging.warn() would have helped immensely.  If this warning / logging IS supposed to be happening, and I just don't see them - can you illuminate me what the appropriate debugging strategy is when you receive a ```<dict>``` instead of your expected class?

Thanks!

P

Theelx:
Closes #190 
Closes #313 
Also beautify code using isort and ssort. I accidentally ran isort because that's part of my normal workflow, so I figured why not go all-in on code cleanup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ TOXCMD += --develop --skip-missing-interpreters
 ifdef multi
     TOXCMD += --parallel $(NUM_JOBS)
     TOXCMD += -e
-    TOXCMD += 'clean,py{27,36,37,38,39},py{27,38,39}-sa{10,11,12,13},py{27,38,39}-libs'
+    TOXCMD += 'clean,py{27,36,37,38,39},py{27,38,39}-sa{12,13},py{27,38,39}-libs'
     # Disable coverage when running in parallel
     TOXCMD += -- --no-cov
 endif
--- a/jsonpickle/__init__.py
+++ b/jsonpickle/__init__.py
@@ -69,17 +69,16 @@ added to JSON::
 """
 from __future__ import absolute_import, division, unicode_literals
 
-from .backend import json
-from .pickler import encode
-from .unpickler import decode
-
 # Export other names not in __all__
 from .backend import JSONBackend  # noqa: F401
-from .version import __version__  # noqa: F401
+from .backend import json
 from .handlers import register  # noqa: F401
 from .handlers import unregister  # noqa: F401
 from .pickler import Pickler  # noqa: F401
+from .pickler import encode
 from .unpickler import Unpickler  # noqa: F401
+from .unpickler import decode
+from .version import __version__  # noqa: F401
 
 __all__ = ('encode', 'decode')
 
--- a/jsonpickle/backend.py
+++ b/jsonpickle/backend.py
@@ -1,7 +1,6 @@
 from __future__ import absolute_import, division, unicode_literals
 
-from .compat import string_types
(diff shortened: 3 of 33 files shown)


Theelx:
Apologies for including code beautification in this PR to everyone looking at the files changed, I didn't realize so much would be changed. The changes from just the adding on_missing are here: https://github.com/jsonpickle/jsonpickle/pull/384/commits/4f3b3dbe778e895fa8c8ff76835a94c4d44cef61

Theelx:
For the record, none of these test failures on 2.7 are able to be reproduced on my local machine when I run tox.

PR #386, opened 2022-05-08
Auto-apply black changes to commits to main

Theelx:
Also remove black from the pytest matrix, since github actions will automatically fix the code and pytest-black was causing 73 deprecation warnings, due to not being updated since October 5th, 2020.

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,0 +1,34 @@
+name: Lint
+
+permissions:
+   contents: write
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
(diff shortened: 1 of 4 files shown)


Theelx:
I've tested the workflow in my fork of jsonpickle, but I had to overwrite the testing commits for a clean pull-request to here. So yeah, it definitely works.

Theelx:
Oh and to clarify, I requested a review from you/davvid in case you have objections to auto-fixing misformatted code using black. I'm not really concerned about confusing new contributors, as this won't run on PRs unless they're merged to main, but rather whether you think this could conflict with the flake8 settings or if you have moral objections to automatically changing code that a human committed.

Theelx:
Merging this, we can always revert later if needed.

davvid:
Thanks for doing this, it's great to having an auto-formatter in place handling this stuff.

Issue #148, opened 2016-07-06
Add ability to decode an object from a dict/json string based on provided class


nyxcalamity:
Would be nice to have some functionality that allows object recreation based on provided class. For example:

```
class Thing(object):
    def __init__(self, name):
        self.name = name

obj = Thing('Awesome')
import jsonpickle
oneway = jsonpickle.encode(obj, unpicklable=False)
result = jsonpickle.decode(oneway, Thing) # or Thing.__class__
assert obj.name == result.name
```

y0rd4nis:
Right now I'm decoding the json string into dict, updating the dict with the corresponding object class, encoding back the updated dict to json string and then calling jsonpickle.decode(...).

>  json_dict = json.loads(json_str)
>  json_dict.update({"py/object": "module.Class"})
> obj = jsonpickle.loads(json.dumps(json_dict))

It works, but it will be nice if the desired class can be provided to jsonpickle.decode(...), like @nyxcalamity says, or specifying the python path to the object class.

ejwessel:
I have a similar problem.

Using the above solution provided by @y0rd4nis I've made the following. Though I don't like it and I'm still open for alternative solutions that are better.
```
from jsonpickle import encode, decode
import json

def to_json(obj):
    return encode(obj)

def from_json(desired_obj, json_str):
    '''
    This essentially seals the py/object info from a passed class
    The big problem with this method is that if the desired_obj has parameters, you'll need pass blanks
    :param desired_obj: instantiated object that we want to
    :param json_str: 
    :return: 
    '''
    object_type = desired_obj.__module__ + "." + desired_obj.__class__.__name__
    json_dict = json.loads(json_str)
    json_dict.update({"py/object":object_type})
    return decode(json.dumps(json_dict))
```

ejwessel:
Here's a better method where you don't have to create a blank object and can only provide the type

```
def from_json2(class_type, json_str):
    type_dict_str = to_json(class_type)
    json_type_dict = json.loads(type_dict_str)
    json_dict = json.loads(json_str)
    json_dict.update({"py/object": json_type_dict["py/type"]})
    return decode(json.dumps(json_dict))
```

```
newObj2 = from_json2(ClassName, json_str)
print isinstance(newObj2, ClassName) -----> True
```

ejwessel:
Regardless of the solutions above there might be a correct way instead of the work arounds proposed, but I was came up short on solutions. 

[...]

PR #392, opened 2022-06-17
Allow non-fully qualified paths in classes argument to deserializer

Theelx:
`unpickler.loadclass` now checks to try and return the class even if the fully-qualified class name doesn't match anything in the classes dict. We do this by allowing users to pass a non-fully qualified class name to decode.


Theelx:
Fixes #148

Issue #395, opened 2022-06-21
Test failures with Python 3.11


hroncok:
Hello. We are trying to rebuild jsonpickle in Fedora with Python 3.11.0b3. Currently, there are various test failures in our build environment, but I can reproduce many with plan tox. To avoid dealing with compilation issues of scipy and pandas, I've reduced the tests like this:

```patch
diff --git a/setup.cfg b/setup.cfg
index f068a00..7712763 100644
--- a/setup.cfg
+++ b/setup.cfg
@@ -57,14 +57,7 @@ testing =
     pytest-cov
 
     # local
-    ecdsa
-    feedparser
-    gmpy2
-    numpy
-    pandas
-    pymongo
-    scikit-learn
-    sqlalchemy
+    pytz
 ```

And run `tox -e py311 -- -k 'not (test_thing_with_module or test_thing_with_submodule)' tests 2>&1 | tee tox_output`.

(test_thing_with_module and test_thing_with_submodule hang for me when reproducing like this, so I've deselected them)

This is the output:

```
WARNING: The wheel package is not available.
py311 develop-inst-noop: .../jsonpickle
py311 installed: attrs==21.4.0,black==22.3.0,click==8.1.3,coverage==6.4.1,docutils==0.18.1,flake8==4.0.1,importlib-metadata==4.11.4,iniconfig==1.1.1,-e git+ssh://git@github.com/jsonpickle/jsonpickle.git@c2fdc01887d6173fdd46e0d322832b906ee80e22#egg=jsonpickle,mccabe==0.6.1,mypy-extensions==0.4.3,packaging==21.3,pathspec==0.9.0,pep517==0.12.0,platformdirs==2.5.2,pluggy==1.0.0,py==1.11.0,pycodestyle==2.8.0,pyflakes==2.4.0,pyparsing==3.0.9,pytest==7.1.2,pytest-black==0.3.12,pytest-black-multipy==1.0.1,pytest-checkdocs==2.7.1,pytest-cov==3.0.0,pytest-flake8==1.1.1,pytz==2022.1,toml==0.10.2,tomli==2.0.1,zipp==3.8.0
py311 run-test-pre: PYTHONHASHSEED='573023578'
py311 run-test: commands[0] | pytest -k 'not (test_thing_with_module or test_thing_with_submodule)' tests
============================= test session starts ==============================
platform linux -- Python 3.11.0b3, pytest-7.1.2, pluggy-1.0.0
[...]

PR #396, opened 2022-06-22
Ignore Python 3.11's default __getstate__ implementation

encukou:
Fixes: #395
*Python 3.11 adds a default `__getstate__` to all objects, which breaks jsonpickle. See the issue for details.*

If jsonpickle wants to keep its current behavior (and not have some objects that serialized as `py/reduce in 3.10 turn into `py/object` with `py/state`), the pickler can ignore the default method.

With this the tests should pass on 3.11. (I didn't try with *all* test dependencies, though.)

Please consider this a draft/suggestion. I'm not a pickling expert by any means.

### DIFF
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -476,8 +476,12 @@ class Pickler(object):
 
         # Support objects with __getstate__(); this ensures that
         # both __setstate__() and __getstate__() are implemented
-        has_getstate = hasattr(obj, '__getstate__')
+        has_own_getstate = (
+            hasattr(type(obj), '__getstate__')
+            and type(obj).__getstate__ is not getattr(object, '__getstate__', None)
+        )
         # not using has_method since __getstate__() is handled separately below
+        # Note: on Python 3.11+, all objects have __getstate__.
 
         if has_class:
             cls = obj.__class__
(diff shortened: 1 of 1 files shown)


AdamWill:
This makes all the tests in @hroncok 's reduced "no-compilation" set pass at least, confirmed. I'll try a mock package build to see if *all* tests pass. Not sure if this is the correct/best fix, but it does seem to do the job at least. Thanks.

Theelx:
Ok, it looks good to me. When I pickle an object from 2.2.0 and unpickle in your fork, it returns the correct deserialized object. So, there should be no need for a helper function. I'm all good with this, just waiting on a second opinion from @davvid.

AdamWill:
Yeah, that makes sense based on what the change does, I wouldn't expect there to be compatibility issues. This only changes pickler behaviour, and it should only change it such that on 3.11 it now behaves the way it used to with 3.10.

Theelx:
I'm going to merge this beause I like the implementation, and davvid can revert if needed. When do you guys need 3.11 support by @encukou @AdamWill @hroncok? There's some other work I'd like to get in before I release 3.0.0, but if y'all need this patch in a public release soon then that can be arranged (but it'd probably be 2.2.1 or 2.3.0).

davvid:
Thank you so much for making this happen. Really appreciate the attention to new Python versions.

AdamWill:
@Theelx it's not urgent to get it in a public release, I've already backported the patch for our downstream package build. As long as it's on track for the next upstream release, that's all good. Thanks a lot!

vstinner:
Nice fix :-)

PR #397, opened 2022-07-12
Allow frozen dataclasses to be deserialized

### LINKED ISSUE #240: jsonpickle.decode does not work for immutable fields in dataclasses
When playing around with the new `dataclass` functionality in python 3.7 I found that jsonpickle does not play with immutable `field`s.

Minimal example:

    @dataclass(frozen=True)
    class c():
        x:str

then 

    jsonpickle.decode(jsonpickle.encode(c(1))) 

throws an `FrozenInstanceError` during the decode step.

Theelx:
Fixes #240

### DIFF
--- a/jsonpickle/ext/pandas.py
+++ b/jsonpickle/ext/pandas.py
@@ -161,6 +161,11 @@ class PandasIndexHandler(BaseHandler):
         buf, meta = self.pp.restore_pandas(data)
         dtype = meta.get('dtype', None)
         name_bundle = {k: v for k, v in meta.items() if k in {'name', 'names'}}
+        if 'names' in name_bundle:
+            # pandas won't accept the names arg in a future version
+            names = name_bundle.pop('names')
+            # name arg must be hashable so cast to tuple first
+            name_bundle['name'] = tuple(names)
         idx = self.index_constructor(decode(buf), dtype=dtype, **name_bundle)
         return idx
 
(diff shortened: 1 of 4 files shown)


jsonpickle/ext/pandas.py
```diff
@@ -161,6 +161,11 @@ def restore(self, data):
         buf, meta = self.pp.restore_pandas(data)
         dtype = meta.get('dtype', None)
         name_bundle = {k: v for k, v in meta.items() if k in {'name', 'names'}}
+        if 'names' in name_bundle:
+            # pandas won't accept the names arg in a future version
+            names = name_bundle.pop('names')
+            # name arg must be hashable so cast to tuple first
+            name_bundle['name'] = tuple(names)
```
davvid, line 168: Thanks for these updates, this is really good stuff.

I'm not sure if this would help at all, but this block of code made me wonder, "can we maybe save one dict lookup by trying to `pop()` unconditionally and then catching the KeyError there?

eg...

```
try:
    names = name_bundle.pop('names')
    ...
except KeyError:
    pass  # 'names' is not present in name_bundle
```

Pretty trivial and this is certainly fine as-is, but sometimes every little dict lookup counts.

Theelx: Oh whoops, this wasn't supposed to be in this commit, I guess I forgot to clean my work tree! Anyway, sure, I can benchmark that change. `try/except` blocks aren't "zero-cost" until CPython 3.11 though, so if it has to catch an error in a lower version that could slow it down a bit.

PR #400, opened 2022-07-27
Pickling a function returns None as the module

effiroz:
Pickling the readline function of io.BytesIO returns a None module instead of _io

```
>>> jsonpickle.encode(BytesIO(b'bytes').readline)
'{"py/function": "None.BytesIO.readline"}'
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,7 @@ v3.0.0
       into different classes. (#148) (+392)
     * Fix bug with deserializing `numpy.poly1d`. (#391)
     * Allow frozen dataclasses to be deserialized. (#240)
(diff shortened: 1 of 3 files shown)


stefan6419846:
It probably makes sense to write a test for this as well.

Theelx:
I'm on vacation right now (on my phone), so I can't check the code quality immediately. Hopefully in about 6-7 hours I'll be back at the rental house, then I can check over this PR on my laptop. I'll approve the workflow for now.

Theelx:
Would you mind also writing an entry for this under 3.0.0 in CHANGES.md? I see you followed the pattern for writing the test, so the CHANGES format should be obvious, thank you so much for the test!

PR #402, opened 2022-08-09
Add python 3.11 to testing matrix

### LINKED ISSUE #401: [TODO] Add Python 3.11 Testing
This is an issue to inform anyone watching this repository that because SciPy released its version with 3.11 support a week or two ago, jsonpickle should be ready to integrate 3.11 support into the testing matrix without extensive workarounds. Hopefully I can get to this within a few days.

Theelx:
Closes #401

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
 
@@ -32,7 +32,7 @@ jobs:
           echo "::set-output name=dir::$(pip cache dir)"
 
       - name: Cache
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         with:
           path: ${{ steps.pip-cache.outputs.dir }}
           key:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,6 +42,11 @@ jobs:
             ${{ matrix.os }}-${{ matrix.python-version }}-v1-
 
       - name: Install dependencies
+        if: matrix.python-version == '3.11-dev'
+        run: |
+          sudo apt-get install -y libopenblas-dev
+
+      - name: Install dependencies
         run: |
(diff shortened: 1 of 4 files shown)


davvid:
Nicely done! Thanks for keeping an eye on the latest python versions. Bummer about the test failures tho.. they seem surprising.. they all seem to bail out pretty early.

Theelx:
The test failures are because flake8 released v5.0 since the last time CI had to run. This introduces compatibility issues and I'm trying to pin it below 5.0 but it isn't working.

Theelx:
Well it seems scipy says they support python 3.11 in 1.9.0 but didn't add wheels for the rc or beta. I'll try and see if I can coax it to build from source, but that seems unlikely.

.github/workflows/test.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11.0-beta.5"]
```
jaraco, line 14: I recommend "N.N-dev" for all versions, as [recommended by the action maintainers](https://github.com/actions/setup-python/issues/213).

pytest.ini
```diff
@@ -1,6 +1,6 @@
 [pytest]
 norecursedirs=dist build .tox .eggs
-addopts=--doctest-modules --flake8 --cov
+addopts=--doctest-modules --cov
```
jaraco, line 3: I suspect you're removing pytest-flake8 because a recent release of flake8 broke the plugin. See jaraco/skeleton for how I've addressed the issue (https://github.com/jaraco/skeleton/commit/c64902b8cafa8062398ef173278a21b042b03a77).

jaraco (approved):

pytest.ini
Theelx: That issue with flake8 breaking it motivated me to drop flake8, however I was planning to drop flake8 eventually anyway because black is more consistent and popular. I also integrated black into GitHub so that any commits that don't follow black style automatically get changed to black style on merge/push.

.github/workflows/test.yml
Theelx: Sounds good, will do!

PR #403, opened 2022-08-11
Remove old features

Theelx:
Removing the bytes/quopri and ref decoding from the unpickler causes list unpickling to be sped up by ~5%.


davvid:
Awesome, this look great!

Theelx:
Thanks! I can't decide whether this PR should replace the other open one or if I should merge the other one first (once it works of course)/

PR #405, opened 2022-08-27
LICENSE: update text so that github detects the 3-clause BSD license

davvid:
The github license checker is strict and wants the license text to
exactly match the text of the BSD 3-clause ("New BSD") license.

Adjust the license text to avoid minor differences that causes github
to classify our license as "other".

This is the 3-clause BSD license.

Signed-off-by: David Aguilar <davvid@gmail.com>

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,29 +1,28 @@
 Copyright (C) 2008 John Paulett (john -at- paulett.org)
 Copyright (C) 2009-2021 David Aguilar (davvid -at- gmail.com)
-All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
-modification, are permitted provided that the following conditions
-are met:
+modification, are permitted provided that the following conditions are met:
 
- 1. Redistributions of source code must retain the above copyright
-    notice, this list of conditions and the following disclaimer.
- 2. Redistributions in binary form must reproduce the above copyright
-    notice, this list of conditions and the following disclaimer in
-    the documentation and/or other materials provided with the
-    distribution.
- 3. The name of the author may not be used to endorse or promote
-    products derived from this software without specific prior
-    written permission.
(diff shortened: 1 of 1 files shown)


davvid:
I noticed that tidelift was using the github license API and github was reporting our license as "other" rather than the bsd-3-clause license. There were some trivial edits to the license text that was causing this.

I *think* this should resolve it. I did trim trailing whitespace as well -- hopefully it's resilient to that, but if not I can restore that to make the license text exactly match.

davvid:
From https://support.tidelift.com/hc/en-us/articles/4406287851924-Debugging-license-problems#how-github-license-scanning-works-0-2

> How it parses
> 
> From the list of licensing files, the parser discards lines that look like copyright statements ("Copyright (c) 1994 Linus Torvalds") and then compares the remaining contents of those files to the full text of known licenses. If the license information is ill-formatted, slightly tweaked, not the full license text (for example, "This project is under the BSD license" rather than the actual BSD license), GitHub will report NOASSERTION. Because of this simplistic (but robust) approach to scanning, another common failure case is a  LICENSE file with more than one license. 


"compares the remaining content", "ill-formatted" and "slightly tweaked" all sound like they could get tripped up by me having trimmed the trailing whitespace. I may just go ahead and amend this commit to retain the trailing whitespace since that might be necessary based on my reading of that explanation.

davvid:
Actually, I'll leave this PR as-is without amending the commit to restore the trailing spaces. I'm comparing against some other repos where bsd-3-clause is properly detected and they also have trailing spaces at the end of lines trimmed, so for now I have hope that github's checker is doing a `str.strip()` of the content lines when comparing.

Theelx (approved):
LGTM, it seems to be a copy/paste of BSD3, which is what we want.

davvid:
Thanks!

Issue #412, opened 2022-10-25
Include py/id attribute in all object declaration to identify univocally each object in a JSON file


AZblo:
We would like to add the py/id attribute next to all objects in the JSON file. 

- In the current implementation no py/id is given to the first instance of an object in the py/id. When encoding/decoding, each new object is given an ID based in their order of appeareance in the json file. 
- While this implementation is perfectly fine for encoding/decoding using jsonpickle; in order to have a more generic implementation, compatible with libraries in other programming languages it is proposed to make appear explicitly the py/id of the first instance of an object in the JSON file. 

**Current implementation**
`{
    "py/obj": "javaTypes.SystemComponent",
    "block": {
        "py/obj": "javaTypes.Block",
        "id": 25,
        "port": {
            "py/obj": "javaTypes.Port",
			"id": 35,
            "name": "puerto2"
        },
        "link": {
            "py/obj": "javaTypes.Link",
			"id": 45,
			"port": {"py/id": 17002}
        }
    },
    "link": {
        "py/id": 3
    },
    "port": {
        "py/id": 2
    }
}`

**Desired implementation**
(note the new attribute py/id appearing next to each py/obj declaration)
`{
    "py/obj": "javaTypes.SystemComponent",
	"py/id": 0,
    "block": {
        "py/obj": "javaTypes.Block",
		"py/id": 1,
        "id": 25,
        "port": {
            "py/obj": "javaTypes.Port",
			"py/id": 2,
			"id": 35,
            "name": "puerto2"
        },
        "link": {
            "py/obj": "javaTypes.Link",
			"py/id": 3,
			"id": 45,
			"port": {"py/id": 17002}
        }
    },
    "link": {
        "py/id": 3
    },
    "port": {
        "py/id": 2
    }
}`

** Advantages **
- Format compatible to Java JackSON library, very useful for exchanging JSON file between Java and Python modules, creating the same object in both ends. 
- Possibility to parse the JSON file unambiguosly, indepently from the order of appeareance of the new object instances.

AZblo:
[https://github.com/AZblo/jsonpickle/tree/hotfix_JackSON_style](url)

Theelx:
[...]

PR #413, opened 2022-10-26
Hotfix jack son style

AZblo:
Hotfix with solution for #412

(I apologize in advance I was not able to run the tests, any help on the error will be much appreciated - when installing numpy in the virtual environment I got: `AttibuteError: fcompiler`)

### DIFF
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -19,777 +19,796 @@ from .compat import numeric_types, string_types
 
 
 def encode(
-    value,
-    unpicklable=True,
-    make_refs=True,
-    keys=False,
-    max_depth=None,
-    reset=True,
-    backend=None,
-    warn=False,
-    context=None,
-    max_iter=None,
-    use_decimal=False,
-    numeric_keys=False,
-    use_base85=False,
-    fail_safe=None,
-    indent=None,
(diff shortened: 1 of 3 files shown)


stefan6419846:
Your changeset is much too large as you replaced all the indentation of four whitespace characters with tabulators. Additionally, having a quick glance at the code, tests are still missing and `pyId` is no valid PEP8 variable/parameter name.

Is the *wheel* package installed inside your virtual environment? Otherwise, *pip* will try to install *numpy* from source, while you can use the existing wheels if you install the *wheel* package beforehand.

Theelx:
Stefan is correct on all of those issues. Please run `black .` and `isort . --profile black` from the command line on the code for proper formatting. You can install them with `pip install black` and `pip install isort`. We also need tests for this, so make some example data and tests, and then run the tests with `pytest .` in the command line of the repo's home directory.

AZblo:
Thank you for your feedback. I will try to run the tests (I am not very experienced yet). 
In the meanwhile, I confirm that the proposed implementation is compatible with Java Jackson library (https://github.com/FasterXML/jackson) - we have been able to fully code and decode a JSON file with object references exchanging it back and forth between Java and Python :)

PR #418, opened 2022-12-01
Speed up unpickling

Theelx:


Theelx:
Also add repeat functionality to pytest. This lets me profile jsonpickle using the test suite rather than a custom script.

PR #421, opened 2022-12-06
Use stdlib datetime.timezone instead of pytz in tests

musicinmybrain:
This removes an implicit test dependency (satisfied via `pandas`) on `pytz`.


davvid:
This looks good to me. I'll let @Theelx merge for another set of eyes.

Theelx:
LGTM! Thanks for helping reduce test dependencies!

PR #433, opened 2022-12-28
Warn if a custom pickling handler returns None

jamesrobson-secondmind:
The docstring for the encode function says
```
    :param warn: If set to True then jsonpickle will warn when it
        returns None for an object which it cannot pickle
        (e.g. file descriptors).
```

However, if you pass in a file descriptor when running python 3.8 or higher the `TextIOHandler` gets used. This means that no warning is raised because there’s no check if a handler has returned None.


Theelx:
Thanks so much for noticing and fixing this!

Issue #442, opened 2023-05-10
When decoding a json string, jsonpickle does not re-initialize the class instance.


MamoonAhmad:
Lets say I have this class 

```
class LineItem:
   def __init__(self):
        self.var1 = 1
```

When I encode this json with JSON pickle it will save all the attributes and class definition. And after encoding, lets say I change the class to following.

```
class LineItem:
   def __init__(self):
        self.var1 = 1
        self.var2 = 2
```
        
Now after changing the line item, if I decode the existing json with JSON pickle. The newly decoded class instance of LineItem will not have attribute var2. So if I try to access it, it will throw an error.

In my opinion, there should be an option to reinitialize the class instance.

Theelx:
I agree that such an option should exist. Hopefully I can get around to implementing it over the summer (I'm finishing up my final exams over the next few days).

That being said, it might be very difficult, and there are a lot of bugs that are a higher priority at the moment. If you'd like it sooner, I'd be happy to review and merge a PR, but I can't promise that I'll be able to do the work myself over the next few months (since I'm taking summer classes and have a full-time research job).

davvid:
Perhaps slightly magical, but I added a test case in #445 that demonstrates how you can accomplish this using existing jsonpickle features.

The basic gist of it is that you can provide a boilerplate `__new__` alongside `self.default_factory = self.__class__` and that'll force jsonpickle into calling your constructor during object initialization.

This does mean that your constructor must function without any arguments, but since you implement `__new__` then in theory you can do all kinds of interception there and rewire things as needed after creating the instance. Not advised, but technically possible.

After construction jsonpickle will fill in any data that was present in the json payload, so if you have json with old attributes those will continue to be present in new instances.

Hopefully this approach works for your use case.

Theelx:
Thanks so much davvid!

PR #445, opened 2023-05-23
object_test: demonstrate use of __new__ + default_factory

davvid:
jsonpickle does not currently have an explicit method to ensure that an object constructor must be run.

Add a test that demonstrates how to use __new__ alongside the existing default_factory functionality to create an object that will always run its constructor during jsonpickle object initialization.

Classes must opt-in to this behavior by providing both self.default_factory and an implementation of __new__ to force jsonpickle into using the user-provided constructor.

Related-to: #442

### DIFF
--- a/tests/object_test.py
+++ b/tests/object_test.py
@@ -176,6 +176,24 @@ class ThingWithClassAsDefaultFactory(collections.defaultdict):
         return self.__class__()
 
 
+class ThingWithDefaultFactoryRegistry:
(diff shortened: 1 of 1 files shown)


Theelx:
This looks great! I'm looking at the online documentation right now though, and I think it might be useful to add a documentation page answering FAQs such as this one, so people don't have to search through Github issues and PRs. However, that's a task for another PR, so I'll just merge this one now.

PR #455, opened 2023-08-05
Fix ref-count when pickling dict subclasses

### LINKED ISSUE #453: Dictionary subclasses with referencing are encoded/decoded incorrectly
```python
import jsonpickle

class MyDict(dict):
    def __init__(self, obj=None):
        self.obj = obj or object()

d = MyDict()
d["inner"] = MyDict(d.obj)

assert d.obj == d["inner"].obj

json = jsonpickle.encode(d)

d2 = jsonpickle.decode(json)
assert d2.obj == d2["inner"].obj
```

The second assertion fails, because while `d2["inner"].obj` is correctly decoded to an `object`, `d2.obj` is somehow set to `d2["inner"].__dict__`!

```python
>>> d2["inner"].obj
<object object at 0x0000022AEAD35700>
>>> d2.obj
{'obj': <object object at 0x0000022AEAD35700>}
```

The encoded JSON is:
```json
{
  "py/object": "__main__.MyDict",
  "inner": {
    "py/object": "__main__.MyDict",
    "__dict__": {
      "obj": {
        "py/object": "builtins.object"
      }
    }
  },
  "__dict__": {
    "obj": {
      "py/id": 2
    }
  }
}
```

GREsau:
Fixes #453

### DIFF
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -775,9 +775,12 @@ class Pickler(object):
 
         # Sub-classes of dict
         if hasattr(obj, '__dict__') and self.unpicklable and obj != obj.__dict__:
-            dict_data = {}
-            self._flatten_dict_obj(obj.__dict__, dict_data)
-            data['__dict__'] = dict_data
+            if self._mkref(obj.__dict__):
+                dict_data = {}
+                self._flatten_dict_obj(obj.__dict__, dict_data)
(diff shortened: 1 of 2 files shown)


Theelx:
Thanks for the contribution! I think this may actually solve some other issues with reference counting being off-by-one, but I'm not sure yet.

PR #456, opened 2023-08-05
Fix depth tracking with list/dict referencing

GREsau:
Currently, the Pickler's depth is incorrectly incremented when flattening lists/dicts that are referenced via `py/id`, and it is not decremented again:

```python
import jsonpickle

l = []
pickler = jsonpickle.Pickler()
print(pickler._depth) # -1

pickler.flatten([])
print(pickler._depth) # -1

pickler.flatten([l])
print(pickler._depth) # -1

pickler.flatten([l, l])
print(pickler._depth) # 0

pickler.flatten([l, l, l, l, l])
print(pickler._depth) # 3
```

With the changes in this PR, all print statements in the above code output the expected value of -1

### DIFF
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -788,7 +788,6 @@ class Pickler(object):
                     self._list_recurse if type(obj) is list else self._flatten_dict_obj
                 )
             else:
-                self._push()
                 return self._getref
 
         # We handle tuples and sets by encoding them in a "(tuple|set)dict"


stefan6419846:
This might should have a corresponding test as well to ensure this behaves correctly and will not break in a future version.

Theelx:
Thank you so much! This PR and #455 are extremely helpful, and will likely fix a lot of issues with `py/id`!  I must agree with Stefan though, a test is necessary for this case. If you'd rather, I can make and push a test, but I want to give you the option first in case you think there need to be additional tests along these lines.

GREsau:
I didn't think a test case for this bug/fix would be very useful, since it's such a specific case. The now-deleted broken line of code was presumably left over from a long-past refactoring, so I feel like writing a test for this would only really be testing that this specific no-longer-existing line of code still doesn't exist 😄 

Bear in mind the `_depth` field doesn't impact the `py/id` values - it's only used for making sure we haven't reached the max depth, so it's actually quite difficult for this bug to actually cause any user-impact, and writing a test to ensure lists/dicts with referencing can be pickled/unpickled technically wouldn't be sufficient to catch the original problem.

It _could_ be worth adding some `_depth == -1` assertions to existing tests as an extra safety net to make sure nothing else like this happens, but again I don't think it's worth just testing lists/dicts, since the only code that treated them differently (at least with respect to the `_depth` value) no longer exists.

Of course if you want to make a test anyway, I certainly won't stop you! 😄

PR #459, opened 2023-08-13
Add depth test

Theelx:
Add test for #456.


Theelx:
I also ran black and ssort on all the jsonpickle code.

PR #462, opened 2023-08-16
readthedocs: specify build.os

davvid:
readthedocs deprecated build.image. We need to specify the OS used by the build these days.

Related-to: https://blog.readthedocs.com/use-build-os-config/


davvid:
I don't have a good way to test this locally so I'm kinda editing this blind. Hopefully this works ;-)

Theelx:
LGTM!

PR #471, opened 2023-11-30
Fix jsonpickle handling of missing modules or attributes

aviramstr:
I think the inner `except` in `loadclass` is a bug. It's causing `loadclass` to skip attributes it can't find along the module path.

For example: `loadclass("numpy.missing")` returns `numpy` instead of failing.

Also, `loadclass("numpy.bla.bla.bla.array")` returns `numpy.array`.

Update:

There is a test named `test_unpickler_on_missing` which should have caught this, but it coincidentally passed anyway. 
The reason is: 
1. The test was decoding an instance of a local class. This shouldn't work, but because of the above bug, `loadclass` returned the test function as the class, instead of `None`.
2. It was then unable to instantiate the class (because it's not the class, it's the function)
3. The code treated this failure as a case of a missing object, and the test passed.

So I fixed the test to use a nonlocal class (Specifically, the `Outer` class defined in the test suite). I also modified it to test missing nested attributes.

The test still failed at this point, seems like it is because of another bug - a missing call to `_process_missing` in the case `loadclass` returns `None`. When adding it, all tests pass.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,13 +1,10 @@
-v3.0.4
-======
-    * Fixed a bug where unpickling a missing class would return a different object
-      instead of ``None``. (+471)
-    * Fixed the handling of missing classes when setting ``on_missing`` to ``warn`` or ``error``. (+471)
-
 v3.0.3
 ======
     * Fixed a bug where pickling some built-in classes (e.g. zoneinfo) 
       could return a ``None`` module. (#447)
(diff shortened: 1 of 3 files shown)


stefan6419846:
You should probably add corresponding test(s) as well.

aviramstr:
Thanks @stefan6419846 , actually there was already a test that should have caught this. See the updated description.

aviramstr:
@Theelx  sure, done.

PR #473, opened 2024-01-08
tests: generate coverage reports using tox

davvid:


davvid:
yay for CI tests. It looks like the older ubuntus don't have tox v4. I'll take a stab at getting the CI tests to install tox v4 before running the tox tests.

.coveragerc
```diff
@@ -1,6 +1,5 @@
 [run]
 omit = .tox/*
-parallel = true
```
Theelx, line 3: Is the parallel option no longer supported? I feel like parallel would provide a good speedup for running tests,

README.rst
Theelx, line 1: Is it possible to move the jsonpickleJS movement to a different PR? While I doubt it'll matter in the future to anyone, it's still nice to keep separate issues in separate PRs.

Theelx (changes requested):
We need a CHANGES.rst entry even for internal changes, but apart from that every suggestion of mine is optional.

.coveragerc
davvid: `parallel = true` interestingly doesn't actually control whether coverage runs in parallel. It makes it write to separate files so that coverage could be generated in parallel by an outer process.

What's odd is that `parallel = true` doesn't actually seem to do anything as far as I can tell. I suspect that `pytest-cov` is in charge and its `--cov-append` option is how they allow for parallelism.

Good point about wanting tox to run in parallel. I've updated this MR to now use `tox run-parallel --parallel-live`.

README.rst
davvid: Yup, I'll move that to another PR. I added changelog entries as well.

I have another PR lined up that replaces `flake8` with `ruff`. I'll stack 'em so that the reordering comes after the pytest + tox related changes.

davvid:
Let me know if I should split out another PR from this one -- 5973b26b2f2fc21abbfa2d3af03fcef244f80cce and the commits leading up to it could be split out into a PR that only handles the python3.12-related changes.

Theelx (approved):
This all looks great to me! Thanks for the edits!

PR #477, opened 2024-02-16
pickler/unpickler: support Cython functions with Cython v3

davvid:


davvid:
I split these changes up into several commits mostly to record the different approaches I encountered along the way. The first ones are kinda hacky but the later ones make things simple again. Let me know if you'd prefer that I squash them all together.

QuLogic:
I believe this patch worked: https://copr.fedorainfracloud.org/coprs/churchyard/pandas2/build/7030817/

davvid:
https://pypi.org/project/jsonpickle/3.0.3/ contains this fix.

PR #479, opened 2024-03-06
Avoid backward-compatibility timezone names

cjwatson:
"US/Eastern" and "Australia/Queensland" are counted as backward-compatibility links, and are no longer shipped in all installations.  Use the corresponding canonical timezone names instead.


davvid:
Thanks for these updates, much appreciated!

davvid:
I'm fixing up the tests in #480. It looks like there was a change in pytest-ruff / ruff that's causing the errors. I'll merge this as soon as I get that resolved by pinning down pytest-ruff.

Issue #422, opened 2022-12-11
AttributeError: 'Int64' object attribute '_type_marker' is read-only


DeveloperAnonymous:
**Python Version:** 3.10.8
**Jsonpickle Version:** 3.0.0
**OS:** Docker, python:3.10-slim (Debian GNU/Linux 11)
**Code:** https://github.com/BattleBanana/BattleBanana/blob/master/dueutil/game/players.py#L481-L488
**Stacktrace:** 
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/discord/client.py", line 409, in _run_event
    await coro(*args, **kwargs)
  File "/usr/src/app/run.py", line 293, in on_message
    await events.on_message_event(message)
  File "/usr/src/app/dueutil/events.py", line 92, in on_message_event
    await message_event(ctx)
  File "/usr/src/app/dueutil/events.py", line 14, in __call__
    if await listener(ctx):
  File "/usr/src/app/dueutil/game/game.py", line 249, in on_message
    player = players.find_player(message.author.id)
  File "/usr/src/app/dueutil/game/players.py", line 472, in find_player
    elif load_player(user_id):
  File "/usr/src/app/dueutil/game/players.py", line 486, in load_player
    loaded_player = jsonpickle.decode(player_data)
  File "/usr/local/lib/python3.10/site-packages/jsonpickle/unpickler.py", line 88, in decode
    return context.restore(data, reset=reset, classes=classes)
  File "/usr/local/lib/python3.10/site-packages/jsonpickle/unpickler.py", line 362, in restore
    value = self._restore(obj)
  File "/usr/local/lib/python3.10/site-packages/jsonpickle/unpickler.py", line 344, in _restore
    return restore(obj)
  File "/usr/local/lib/python3.10/site-packages/jsonpickle/unpickler.py", line 768, in _restore_object
    return self._restore_object_instance(obj, cls, class_name)
  File "/usr/local/lib/python3.10/site-packages/jsonpickle/unpickler.py", line 744, in _restore_object_instance
    instance = self._restore_object_instance_variables(obj, instance)
  File "/usr/local/lib/python3.10/site-packages/jsonpickle/unpickler.py", line 690, in _restore_object_instance_variables
    instance = self._restore_state(obj, instance)
  File "/usr/local/lib/python3.10/site-packages/jsonpickle/unpickler.py", line 649, in _restore_state
    state = self._restore(obj[tags.STATE])
[...]

PR #481, opened 2024-03-10
Fix #478

### LINKED ISSUE #478: Error deserializing django SafeString
Serializing and then deserializing a django SafeString runs into error:
_AttributeError: 'SafeString' object attribute 'capitalize' is read-only_

jsonpickle version: 3.0.3
django version: 4.2.11

Theelx:
This issue was very similar to https://github.com/jsonpickle/jsonpickle/issues/422#issuecomment-1355438316. In that case, the issue was because the attribute is being inherited from a class implemented in C, which can make read-only attributes. The fix I implemented in that case was to skip it for subclasses of int, but since this is happening for subclasses of str too, this will add str to the list of parent classes that can be skipped. Eventually, I'll make a more flexible check that doesn't require us to whitelist types, but that's going to have to be on the TODO list for now.
Note: If this is merged, squash the commits instead of merging all of them.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,12 +5,15 @@ v3.0.3
       could return a ``None`` module. (#447)
     * Fixed a bug where unpickling a missing class would return a different object
       instead of ``None``. (+471)
-    * Fixed the handling of missing classes when setting ``on_missing`` to ``warn`` or ``error``. (+471)
+    * Fixed the handling of missing classes when setting ``on_missing`` to ``warn``
+      or ``error``. (+471)
     * The test suite was made compatible with Python 3.12.
     * The tox configuration was updated to generate code coverage reports.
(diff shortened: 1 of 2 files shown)


davvid (approved):
Nicely done! Thanks for the detailed commit messages.

davvid:
Ahh.. the only fixup might be to start a new `v3.0.4` section in the changelog. I believe I've already tagged and released `v3.0.3`.

Theelx:
> Ahh.. the only fixup might be to start a new `v3.0.4` section in the changelog. I believe I've already tagged and released `v3.0.3`.

Ah whoops, I checked PyPi and it only showed 3.0.2 (either I misread it or it showed a cached page). Anyone, I see it now says 3.0.3, so I'll change this and merge!

PR #482, opened 2024-03-12
tests: enable cache for pytest-ruff >= 0.3.0

davvid:
Newer versions of pytest-ruff require a cache provider.
Remove the no:cacheprovider flags to get pytest-ruff working again.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,8 @@ v3.0.4
 ======
     * Fixed an issue with django.SafeString and other classes inheriting from
       str having read-only attribute errors (#478) (+481)
+    * `pytest-ruff` was pinned to the `0.2.x` versions to defer a breaking
+      change in `ruff>=0.3.0`.
 
 v3.0.3
 ======
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,8 +2,7 @@ v3.0.4
 ======
     * Fixed an issue with django.SafeString and other classes inheriting from
       str having read-only attribute errors (#478) (+481)
-    * `pytest-ruff` was pinned to the `0.2.x` versions to defer a breaking
-      change in `ruff>=0.3.0`.
+    * The test suite was made compatible with `pytest-ruff>=0.3.0`. (+482)
 
 v3.0.3
 ======
--- a/setup.cfg
+++ b/setup.cfg
@@ -48,7 +48,7 @@ testing =
 	pytest-checkdocs >= 1.2.3
 	pytest-cov
 	pytest-enabler >= 1.0.1
-	pytest-ruff < 0.3
+	pytest-ruff >= 0.2.1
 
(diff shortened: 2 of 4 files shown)


davvid:
I'm going to take another look. I think the real error is that we're disabling pytest's cache provider, and that's causing pytest-ruff to fail because the newer versions rely on being able to store and retrieve items from the cache.

I'll update this PR shortly after I make the same change over in https://github.com/git-cola/git-cola/

setup.cfg
```diff
@@ -47,7 +48,7 @@ testing =
 	pytest-checkdocs >= 1.2.3
 	pytest-cov
 	pytest-enabler >= 1.0.1
-	pytest-ruff != 0.3
+	pytest-ruff >= 0.2.1
```
Theelx, line 51: Is there something in 0.2.0 that's broken? I'm an advocate of leaving dependency versions as unrestricted as possible so that end users can decide what works best for them.

CHANGES.rst
```diff
@@ -2,6 +2,7 @@ v3.0.4
 ======
     * Fixed an issue with django.SafeString and other classes inheriting from
       str having read-only attribute errors (#478) (+481)
+    * The test suite was made compatible with `pytest-ruff>=0.3.0`.
```
Theelx, line 5: Can you link the PR number?

tox.ini
```diff
@@ -9,7 +9,7 @@ allowlist_externals =
 passenv =
 	FORCE_COLOR
 commands =
-	python3 -m pytest -p no:cacheprovider --cov --cov-append --cov-report=term-missing jsonpickle tests {posargs}
```
Theelx, line 12: LGTM! I do wonder though, why were we not using a cache provider before?

Theelx (changes requested):
The only thing that really should be fixed is the lack of a PR link in CHANGES.rst, but otherwise looks basically good to merge!

setup.cfg
Theelx: Never mind, I just saw that 0.2.0 was yanked because it's broken. I think supporting <0.2.0 might be a pain because those are basically alpha versions, so this looks good!

tox.ini
davvid: Thanks for the review ~ I'll add a PR link shortly.

The only reason I disabled the cache was to avoid `.pycache*` directories on disk ~ there really wasn't a strong reason otherwise. Not having a cache is actually kinda detrimental for others reasons (e.g. it prevents us from being able to use pytest's `--last-failed` and `--failed-first` options) so there's no downside to enabling it.

I suppose I should probably file a bug against pytest-ruff to make it gracefully handle the case where someone uses that flag, but I'll save that for later.

Theelx (approved):
Looks great. Thanks for the detailed answer about not using a cache provider!

PR #483, opened 2024-03-12
Add de/encode feature-flag for readonly attributes

Theelx:
This PR adds a feature-flag that can be set to True for graceful handling of read-only attributes, and which produces concise output. There are a few BS commits in the history because my local black version was out-of-date, and I had some issues doing merges as I'm still trying to get used to git-cola.

In the future, I'm going to update the documentation for decode and other such functions to be more similar to encode, because I noticed some formatting differences that are bugging me.

### DIFF
--- a/tests/jsonpickle_test.py
+++ b/tests/jsonpickle_test.py
@@ -163,6 +163,13 @@ class MyPropertiesDict:
             and list(self.other_object) == list(other.other_object)
         )
 
+# see issue #478
+class SafeData:
+    __slots__ = ()
+
+class SafeString(str, SafeData):
+    __slots__ = ()
+
 
 def on_missing_callback(class_name):
     # not actually a runtime problem but it doesn't matter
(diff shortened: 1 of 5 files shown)


Theelx:
It seems ruff has taken issue with one of the older tests, and that's why the current tests aren't passing: https://github.com/Theelx/jsonpickle/actions/runs/8255950370/job/22583541695#step:8:485

tests/jsonpickle_test.py
```diff
@@ -990,6 +999,16 @@ def test_depth_tracking(self):
...
 
+    def test_readonly_attrs(self):
+        s = SafeString("test")
+        pickled = jsonpickle.encode(s, handle_readonly=True)
+        pickled_json_dict = jsonpickle.backend.json.loads(pickled)
+        # make sure it's giving concise output by erroring if it includes
+        # a default method which is unnecessary
+        self.assertTrue("join" not in pickled_json_dict)
+        unpickled = jsonpickle.decode(pickled)
+        self.assertEqual(unpickled.__class__, SafeString)
```
davvid, line 1010: Nicely done! The only thing I would suggest adding is `assert unpickled == s` so that we're sure that the `test` string value got restored.

davvid (approved):

Issue #484, opened 2024-03-16
Add a "Usage" section to the README


takanuva15:
Hi, could we add a "Usage" section to the README with just a short example of serializing and deserializing a basic python object? This will make it faster to get started with this library instead of having to jump over to the docs for python pickle to understand the correct way to call this library's methods

davvid:
How do we know people will read this section if they didn't already read the intro section that links to the jsonpickle documentation?

That documentation already has a usage example and we shouldn't duplicate all of that stuff in the readme.

I'm not 100% sure that this will solve any perceived issues.

I would actually go in the other direction -- I would recommend slimming the README down and making it link to the documentation instead so that there's less duplication of effort and reading material.

> This will make it faster to get started with this library instead of having to jump over to the docs for python pickle 

We link to jsonpickle's documentation from the README. You don't need to read the python pickle documentation.

I'm not convinced that lazy readers is a problem that can be solved by... adding more reading material.

I'd be happy review a PR that updates the README with a Usage section that links to https://jsonpickle.github.io/#jsonpickle-usage though.

Theelx:
Actually, I think Takanuva has a point, @davvid, but not for the reason they specified. When I google "jsonpickle" or " "jsonpickle docs", the first result I get is our outdated documentation from version 1.4.3. The GitHub page isn't even the in the first 3 results for either search (for me). Until that page is shut down, I think it may be useful to combine a short usage example on the GitHub with a slimmed-down README, so that anyone who happens to land on the GitHub after seeing the outdated docs notices that the library has changed significantly and how to get started.

![Screenshot from 2024-03-17 17-00-14](https://github.com/jsonpickle/jsonpickle/assets/43764914/2e97f143-25d2-444c-8484-fe8456e047ef)

Theelx:
Also, I will say that when I'm under time pressure, I do often check the GitHub README instead of the documentation page, especially when I'm looking among alternatives for a library that fulfills a purpose. I can gather examples of the many python libraries that include a short Usage section, with ~2-5 lines of basic code in their GitHub READMEs, because there's a good reason it's not uncommon for libraries to include such a section.
> lazy readers
[...]

PR #485, opened 2024-03-17
Slightly Slim Down README

Theelx:
This PR removes and combines some old/outdated parts of the README. For example, it removed the jsonpickleJS section, since the JS version hasn't been updated in 10 years (it's probably very broken right now). It also combines the numpy/pandas sections, since those had a decent bit of duplicated text. Lastly, it removes the section detailing how to install jsonpickle if you have the files locally checked out, since it used a deprecated "setup.py develop" install method, which python warns about on 3.11. This last change might be better with a replacement such as "pip install ." when cd-ed into the directory, but I decided against that because I don't believe that works on Windows.

This is a first step towards getting the README in a readable manner in order to implement #484 in a clean way.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -124,12 +124,6 @@ GPG Signing
 ===========
 Unfortunately, while versions of jsonpickle before 3.0.1 should still be signed, GPG signing support was removed from PyPi (https://blog.pypi.org/posts/2023-05-23-removing-pgp/) back in May 2023.
 
-jsonpickleJS
-============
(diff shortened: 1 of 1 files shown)


davvid:
Nice!!

davvid (approved):
This looks great! Thanks again.

I believe `pip install --editable .` should work these days but I haven't tested that on Windows to verify. I do use that on Linux / macOS and things work as expected. The PR is fine as-is, though, since we're removing this section.

PR #487, opened 2024-03-19
Adjust documentation formatting

Theelx:
This PR adds parameter tags to the decode part of the docs, and orders the encode docs to reflect the ordering of the input. Davvid, can you write a short description of what the `reset` and `context` args do, and I can add that to this PR?

### DIFF
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -55,10 +55,6 @@ def encode(
             )
 
         before your pickling code.
-    :param max_depth: If set to a non-negative integer then jsonpickle will
-        not recurse deeper than 'max_depth' steps into the object.  Anything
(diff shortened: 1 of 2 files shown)


davvid:
```
:reset: Custom pickle handlers that use the `Pickler.flatten` method or
    `jsonpickle.encode` function must call `encode` with `reset=False`
    in order to retain object references during pickling.

    This flag is not typically used outside of a custom handler or
    `__getstate__` implementation.


:context: Supply a pre-built Pickler or Unpickler object to use for the
    `jsonpickle.encode/decode` machinery instead of creating a new instance.
    The `context` represents the currently active Pickler / Unpickler when
    custom handlers are invoked by jsonpickle.
```

davvid (approved):

PR #489, opened 2024-04-09
numpy: forward constructor arguments through register_handlers()

davvid:
Expose the NumpyNDArrayHandler constructor arguments to the user-facing `def register_handlers()`. This allows users to opt-out of warnings and customize ndarray parameters.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+v3.1.0
+======
+    * `jsonpickle.ext.numpy.register_handlers` now provides options that are forwarded
+      to the `NumpyNDArrayHandler` constructor. (+489)
+
 v3.0.4
 ======
     * Fixed an issue with django.SafeString and other classes inheriting from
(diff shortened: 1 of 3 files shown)


davvid:
This changes the public API (in a non-breaking way) so this is worth reviewing.

This might be considered a new feature, so if we need to bump the minor version we should probably tag the current state and release the next patch version before merging these changes.

We're exposing a few new keyword arguments and the names of the arguments will be part of the public API so that's definitely worth a look.

CHANGES.rst
```diff
@@ -6,6 +6,8 @@ v3.0.4
     * A `garden.yaml` file was added for use with the
       `garden <https://crates.io/crates/garden-tools>_` command runner. (+486)
     * The test suite was updated to avoid deprecated SQLALchemy APIs.
+    * `jsonpickle.ext.numpy.register_handlers` now provides options that are forwarded
```
Theelx, line 9: This is definitely an API addition, so this should be v3.1.0.

Theelx (approved):
Everything looks good to me, it's a pretty simple change but it should be 3.1.0 nonetheless.

CHANGES.rst
davvid: Excellent. I'll go ahead and start prepping the v3.0.4 release before merging this in.

PR #491, opened 2024-04-11
jsonpickle: add support for top-level tuples with cyclical references

davvid:
Tuples were not previously part of the referencing machinery. This made it so that references to tuples would not get restored -- copies were getting created instead.

Leverage the "mutuple" package to make it so that jsonpickle can edit tuples after they are created. This allows us to replace proxies in tuples similarly to how lists are updated when proxies are replaced.

### DIFF
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -11,6 +11,11 @@ import sys
 import warnings
 from itertools import chain, islice
 
+try:
+    import mutuple
+except ImportError:
+    mutuple = None
+
 from . import compat, handlers, tags, util
 from .backend import json
 from .compat import numeric_types, string_types
@@ -299,11 +304,13 @@ class Pickler(object):
         Return True if this object is new and was assigned
         a new ID. Otherwise return False.
         """
-        objid = id(obj)
-        is_new = objid not in self._objs
-        if is_new:
-            new_id = len(self._objs)
-            self._objs[objid] = new_id
+        obj_id = id(obj)
+        try:
+            _ = self._objs[obj_id]
+            is_new = False
+        except KeyError:
+            self._objs[obj_id] = len(self._objs)
+            is_new = True
         return is_new
 
     def _mkref(self, obj):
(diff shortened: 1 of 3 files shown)


davvid:
I was experimenting a bit and I ended up creating a package that lets us edit tuples.

This is pretty cool because it lets us do stuff that we previously couldn't do in jsonpickle but it's also kinda wild.

I figured this was worth sharing since it might be interesting to folks here.

Here's the package that this leverages:

https://github.com/davvid/mutuple

https://pypi.org/project/mutuple/


The test case that this adjusts demonstrates the new capabilities. It's fully opt-in, but it does change the on-the-wire json representation because tuples are now included in the referencing machinery.

Fun stuff :smile_cat:

davvid:
Fun as it may be, this is definitely a bad idea since it can results in impossible situations involving dicts and sets. It's kinda fun how this breaks the python interpreter and creates tuples that can't be removed from a set or a dict's keys.

This was just for fun so I'm definitely not merging this, but it is curious because the test case does demonstrate a limitation in the current referencing machinery.

I don't now if I've written up my thoughts on how we could solve these kinda situations.. maybe it's in a comment in an issue.. I'll have to dig that up. One idea I considered was changing traversal so that it stores everything in a flat "object store" and serializing inside-out instead of outside-in (or possibly two-pass) so that we can fill in cyclical references after the inner child objects have already been created. But that's pure speculation/intuition and I'm not sure if it would actually work in practice.

The main downside to that kinda approach is that the json representation would end up being very different and not similar to the somewhat "intuitive" structure that's currently encoded today, making it much different from plain json even for simple types like lists and dicts.

In that respect, we might consider jsonpickle "good enough" since it already handles some pretty advanced cases.

davvid:
```python
>>> import mutuple
>>> t = (1, 2, 3)
>>> d = dict()
>>> s = set()

>>> d[t] = 123
>>> s.add(t)

>>> mutuple.setitem(t, 0, 303)
>>> t
(303, 2, 3)

>>> d[t] = 30323
>>> d
{(303, 2, 3): 123, (303, 2, 3): 30323}
>>> d.pop((303, 2, 3))
30323
>>> d
{(303, 2, 3): 123}
>>> d.pop((303, 2, 3))
KeyError: (303, 2, 3)

>>> d.pop((1, 2, 3))  # original t
KeyError: (1, 2, 3)

>>> s
{(303, 2, 3)}
>>> s.add(t)
>>> s
{(303, 2, 3), (303, 2, 3)}
>>> s.remove(t)
>>> s
{(303, 2, 3)}
>>> s.remove(t)
KeyError: (303, 2, 3)

>>> s.remove((1, 2, 3))  # original t
KeyError: (1, 2, 3)
```

:joy_cat:

Theelx:
This is hecking hilarious, and I love the tomfoolery going on!

PR #494, opened 2024-04-19
classes should support replacing with dict

Squadrick:
Without this change, passing a dict in `.decode()` will completely
ignore the key, and just substitute the value directly.

Consider the serialized data (`json_str`):
```
{
  "x": "value",
  "py/object": "older.pkg.Class",
}
```

If `Class` is now moved to `newer.pkg`, I'd expect to use the API like:

```
jsonpickle.decode(json_str, classes={"older.pkg.Class": newer.pkg.Class})
```

Currently the above API doesn't work, it will populate `self._classes` as:
```
{"newer.pkg.Class": new.pkg.Class}
```

The `"older.pkg.Class"` is completely ignored, even though `loadclass` is
implemented to support this case of class replacement.

### DIFF
--- a/jsonpickle/unpickler.py
+++ b/jsonpickle/unpickler.py
@@ -390,8 +390,8 @@ class Unpickler(object):
             for cls in classes:
                 self.register_classes(cls)
         elif isinstance(classes, dict):
-            for cls in classes.values():
-                self.register_classes(cls)
+            for cls in classes:
+                self._classes[cls] = classes[cls]
         else:
             self._classes[util.importable_name(classes)] = classes
 
--- a/jsonpickle/unpickler.py
+++ b/jsonpickle/unpickler.py
@@ -390,8 +390,7 @@ class Unpickler(object):
             for cls in classes:
                 self.register_classes(cls)
         elif isinstance(classes, dict):
-            for cls in classes:
-                self._classes[cls] = classes[cls]
+            self._classes.update(classes)
         else:
             self._classes[util.importable_name(classes)] = classes
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,9 @@ v3.1.0
 ======
     * `jsonpickle.ext.numpy.register_handlers` now provides options that are forwarded
       to the `NumpyNDArrayHandler` constructor. (+489)
+    * Fix bug of not handling ``classes`` argument to `jsonpickle.decode`
+      being a dict. Previously, the keys were ignored and only values were
+      used. (+494)
 
 v3.0.4
 ======
--- a/tests/jsonpickle_test.py
+++ b/tests/jsonpickle_test.py
@@ -571,10 +571,9 @@ class PicklingTestCase(unittest.TestCase):
         obj = MyPropertiesSlots()
         encoded = jsonpickle.encode(obj)
         # MyPropertiesSlots and MyPropertiesDict have compatible eq methods
-        self.assertEqual(
-            obj,
-            jsonpickle.decode(encoded, classes={"MyPropertiesSlots": MyPropertiesDict}),
-        )
+        decoded = jsonpickle.decode(encoded, classes={"MyPropertiesSlots": MyPropertiesDict})
+        self.assertIsInstance(decoded, MyPropertiesDict)
+        self.assertEqual(obj, decoded)
 
     def test_warnings(self):
         data = os.fdopen(os.pipe()[0])


jsonpickle/unpickler.py
```diff
@@ -390,8 +390,8 @@ def register_classes(self, classes):
             for cls in classes:
                 self.register_classes(cls)
         elif isinstance(classes, dict):
-            for cls in classes.values():
-                self.register_classes(cls)
+            for cls in classes:
+                self._classes[cls] = classes[cls]
```
davvid, line 394: We can save a dict lookup here by uisng `for cls, handler in classes.items(): self._classes[cls] = handler`.

Once it's written that way it becomes clearer that we don't even need to loop. It seems like we can use `.update()` instead, which is faster too -> `self._classes.update(classes)`

Make sure to run the tests. I just enabled the actions so we'll see the results from the CI tests shortly.

If we do take this then we'll at minimum need a test case to make sure that it's working as advertised. Similarly, that test will make sure I don't accidentally break this in the future =)

Thanks for the patch, this does sound like a useful feature at first glance.

davvid (changes requested):
If @Theelx agrees I'm down to merge this once a test is added.

@Squadrick if you could kindly add a note to the changelog (`CHANGES.rst`) that'd be helpful as well. `v3.1.0` is still in an unreleased alpha state so it'd go under that section.

jsonpickle/unpickler.py
davvid: Once we see the test it might become clearer that the API needs adjustment. It's kinda interesting that we only looked at the keys.. it's almost like we left the door open to this but hadn't implemented this edge case.

I'm thinking that might actually want to use something like `for key, value in classes.items(): self._classes[util.importable_name(key)] = value` as the implementation.

Likewise, we wouldn't want users to have to stringify things themselves, as that's exposing a bit too much of our internal implementation to the outside world. It's be better to have the API take class objects in both the keys and values, i.e.

```python
jsonpickle.decode(json_str, classes={older.pkg.Class: newer.pkg.Class})
```

In other words, we'd expect the user to pass the class direclty in the key rather than the quoted importable name. That way there's less chance for errors.

davvid: Last sug, I promise (for now) -- I do generally think that passing the classes directly is better.

But, I can imagine scenarios where `older.pkg.Class` is not importable, and we still want users to be able to remap the json data they have around without having to jump through too many hoops.

In that case, we should check if the key is a string and if it is then we can use it as-is. If it's not a string (e.g. it's a class/type object) then we'll call importable_name on it.

That seems like the best of both worlds. It'll be ergonomic for the common case and it'll also let users supply the string names directly.

Theelx: I think @davvid is correct with the suggestion of using ``self._classes.update(classes)``. His assessment that allowing both strings and class objects to work is ideal behavior, but I think that's work for another PR, as I generally take the view that it's best to push out bug fixes first and then add ideal behavior later, once there's enough developer time to go around.

Theelx (changes requested):
I concur with davvid's request for changes, and I think a test and changelog entry would be nice before this PR is merged. If it's preferable for y'all though, I can probably implement the test and changelog entry after finals end for me in mid-May.

jsonpickle/unpickler.py
Squadrick: Agreed on both counts. In this PR, I've fixed the bug of not updating `self._classes` correctly, but it only supports string keys in the dict (concur with @Theelx).

I've made the change of using `.update()` directly.

Squadrick:
@davvid @Theelx Thanks for the review, pushed commits addressing:
1. Using `.update()` directly.
2. Adding a test: There was already a test written, but it was missing an important type check assertion. Added the check now, which fails without this change.
3. Added to changelog.

Squadrick:
Here's a PR for supporting class type as keys in the classes dict: https://github.com/Squadrick/jsonpickle/pull/1

Theelx:
This looks great to me! Thanks for the PR supporting class types as keys too! That PR will be ready to merge as soon as another changelog entry is added for being able to use types as keys :)

Theelx (approved):

Squadrick:
@Theelx Thanks for the review. Regarding the other PR: Let's merge this PR first, and I can raise a diff PR to this upstream repo; it is currently raised in my fork.

davvid (approved):

davvid:
I went ahead and merged both branches. I wanted to get the 2nd merge in there sooner rather than later since it could have changed behavior for some edge cases. Specifically, if someone was previously passing in a dict of `{cls: cls}` then the original version would have allowed it, and calling update on the dict directly would have changed the semantics.

Now we're all set. I'll let this cook in git for a bit. We can plan to tag the next feature release towards the end of May to give @Theelx a chance to adjust things as needed before the next release is tagged.

Heads-up that I ended up changing the dict comprehension into a generator expression to avoid creating an intermediate dict in 29a5584d80d0e0be1e612ab6771360acdf33242e.

 Thanks @Squadrick and @Theelx!

PR #501, opened 2024-05-25
Fix reference tracking when using nested dictionaries/objects under `py/state`

### LINKED ISSUE #500: Nested dictionaries in `py/state` incorrectly restored when tracking object references
Repro:
```python
import jsonpickle


class MyClass:
    def __init__(self, o):
        self.o = o

    def __getstate__(self):
        return self.__dict__


i1 = MyClass({"a": "a"})
i2 = MyClass({"b": "b"})
original = [i1, i1, i2, i2]

assert original[0] is original[1]  # ok
assert original[2] is original[3]  # ok

json = jsonpickle.encode(original)
# Correctly encoded as:
# [
#   {"py/object": "__main__.MyClass", "py/state": {"o": {"a": "a"}}},
#   {"py/id": 1},
#   {"py/object": "__main__.MyClass", "py/state": {"o": {"b": "b"}}},
#   {"py/id": 4}
# ]

decoded = jsonpickle.decode(json)
assert decoded[0] is decoded[1]  # ok
assert decoded[2] is decoded[3]  # fail
```

The assertion fails because `decoded[3]` is incorrectly set to the dictionary `{'a': 'a'}`, instead of the second instance of `MyClass`.

I think this only occurs for types that define `__getstate__` without also defining `__setstate__`.

This occurs because `Unpickler._restore_state` will first restore the dict under `py/state`, which tracks references of any objects under it. But then it goes into `_restore_from_dict`, which restores the same dict _again_, tracking the references of any objects under it _again_.

GREsau:
Fixes #500

This PR causes a dict returned from `__getstate__` (and stored in `py/state`) to no longer be restored twice, which previously caused any objects in it to be incorrectly duplicated in the Unpickler's list of object instances.

### DIFF
--- a/jsonpickle/unpickler.py
+++ b/jsonpickle/unpickler.py
@@ -607,23 +607,26 @@ class Unpickler(object):
 
         return restore_key
 
-    def _restore_from_dict(self, obj, instance, ignorereserved=True):
+    def _restore_from_dict(self, obj, instance, restore_dict_items=True):
         restore_key = self._restore_key_fn()
         method = _obj_setattr
         deferred = {}
 
         for k, v in util.items(obj):
             # ignore the reserved attribute
-            if ignorereserved and k in tags.RESERVED:
+            if restore_dict_items and k in tags.RESERVED:
                 continue
             if isinstance(k, numeric_types):
                 str_k = k.__str__()
             else:
                 str_k = k
             self._namestack.append(str_k)
-            k = restore_key(k)
-            # step into the namespace
-            value = self._restore(v)
+            if restore_dict_items:
+                k = restore_key(k)
+                # step into the namespace
+                value = self._restore(v)
+            else:
+                value = v
             if util.is_noncomplex(instance) or util.is_dictionary_subclass(instance):
                 try:
                     if k == '__dict__':
@@ -690,12 +693,12 @@ class Unpickler(object):
             # implements described default handling
             # of state for object with instance dict
             # and no slots
-            instance = self._restore_from_dict(state, instance, ignorereserved=False)
+            instance = self._restore_from_dict(state, instance, restore_dict_items=False)
         elif has_slots:
-            instance = self._restore_from_dict(state[1], instance, ignorereserved=False)
+            instance = self._restore_from_dict(state[1], instance, restore_dict_items=False)
             if has_slots_and_dict:
                 instance = self._restore_from_dict(
-                    state[0], instance, ignorereserved=False
+                    state[0], instance, restore_dict_items=False
                 )
         elif not hasattr(instance, '__getnewargs__') and not hasattr(
             instance, '__getnewargs_ex__'
--- a/tests/jsonpickle_test.py
+++ b/tests/jsonpickle_test.py
@@ -1102,6 +1102,9 @@ class PickleProtocol2GetStateDict(PickleProtocol2Thing):
     def __getstate__(self):
         return {'magic': True}
 
+class PickleProtocol2GetStateNestedDict(PickleProtocol2Thing):
+    def __getstate__(self):
+        return {'nested': {'magic': True}}
 
 class PickleProtocol2GetStateSlots(PickleProtocol2Thing):
     def __getstate__(self):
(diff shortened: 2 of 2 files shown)


jsonpickle/unpickler.py
```diff
@@ -607,23 +607,26 @@ def restore_key(key):
...
             self._namestack.append(str_k)
-            k = restore_key(k)
-            # step into the namespace
-            value = self._restore(v)
+            if restore_dict_items:
+                k = restore_key(k)
+                # step into the namespace
+                value = self._restore(v)
+            else:
+                value = v
```
davvid, line 629: It looks like we're no longer calling `k = restore_key(k)` in the `else` block here. I haven't dug in deep enough to know whether or not this is an oversight.

Do all the tests still pass if we keep that?

davvid:
Thanks for this enhancement. I'll wait until @Theelx has a chance to review. Also, we're going to want to tag and release v3.1.0 before merging this (unless Theelx thinks otherwise).

Let me know what you think about the `restore_key()` thing and then we can start planning the next release and subsequent merge.

jsonpickle/unpickler.py
GREsau: That is intentional to prevent double-decoding of dict keys - e.g. otherwise this would error:

```python
import jsonpickle

class Foo:
    def __init__(self):
        setattr(self, 'json://123', 'test')
    def __getstate__(self):
        return self.__dict__

f = Foo()
encoded = jsonpickle.encode(f, keys=True)
f2 = jsonpickle.decode(encoded, keys=True)
```

Theelx: This looks good to me!

tests/jsonpickle_test.py
```diff
@@ -1555,6 +1558,20 @@ def test_restore_dict_state(self):
         decoded = jsonpickle.decode(encoded)
         self.assertTrue(decoded.magic)
 
+    def test_restore_nested_dict_state_with_references_preserved(self):
```
Theelx, line 1561: Thanks for adding a test case!

Theelx (approved):
Looks fine! Sorry for not getting around to this sooner, I started a new job a few days ago, and things have been hectic.

Theelx:
> Thanks for this enhancement. I'll wait until @Theelx has a chance to review. Also, we're going to want to tag and release v3.1.0 before merging this (unless Theelx thinks otherwise).
> 
> Let me know what you think about the `restore_key()` thing and then we can start planning the next release and subsequent merge.

This seems like something that we'd want in 3.1.0 rather than 3.1.1 (i.e. sooner rather than later). I think this could fix at least a few other open issues, so delaying it possibly a few weeks/months until 3.1.1 might not be the best. Also, if it does end up somehow accidentally breaking someone's code (unlikely), then we could still revert it and quickly release a 3.1.1 until we can figure out how to resolve it.

jsonpickle/unpickler.py
```diff
@@ -607,23 +607,26 @@ def restore_key(key):
 
         return restore_key
 
-    def _restore_from_dict(self, obj, instance, ignorereserved=True):
```
Theelx, line 610: Wait actually upon closer inspection, even though the method is marked private (with the single underscore), it would be best not to entirely remove this parameter unless we're in a major release (even renaming it can cause issues). Can you keep this parameter and add another one for restore_dict_items?

Theelx (changes requested):
Just took another look at the code, some minor changes would be useful to preserve compatibility and add a changelog entry.

davvid:
> This seems like something that we'd want in 3.1.0 rather than 3.1.1 (i.e. sooner rather than later).
> ...
> Also, if this PR does break something, I think it's better to have it in a minor release rather than a bugfix one

Out of extra caution I tagged v3.1.0 before this merge and I'll tag v3.2.0 now with just this merge in it after writing up a changelog entry for this one.

I should have read my older emails first ;-) Thanks for reviewing!

jsonpickle/unpickler.py
davvid: Shoot I missed this note. I'll tag v3.2.1 with an update that restores `ignorereserved` shortly.

Issue #504, opened 2024-05-30
Python 3.13 regressions


musicinmybrain:
It’s very hard to test this in a `virtualenv` due to dependencies that are not ready for Python 3.13. However, we’re testing with patched dependencies in Fedora, and Python 3.13 will soon be the system Python in Fedora Rawhide (the development version).

When we [tried to build jsonpickle with Python 3.13.0b1](https://copr.fedorainfracloud.org/coprs/g/python/python3.13/build/7509658/), we encountered the following regressions related to importing modules while unpickling:

```
=================================== FAILURES ===================================
___________________ [doctest] jsonpickle.unpickler.loadrepr ____________________
275 Returns an instance of the object from the object's repr() string.
276 It involves the dynamic specification of code.
277 
278 >>> obj = loadrepr('datetime/datetime.datetime.now()')
UNEXPECTED EXCEPTION: NameError("name 'datetime' is not defined")
Traceback (most recent call last):
  File "/usr/lib64/python3.13/doctest.py", line 1395, in __run
    exec(compile(example.source, filename, "single",
    ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                 compileflags, True), test.globs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<doctest jsonpickle.unpickler.loadrepr[0]>", line 1, in <module>
  File "/builddir/build/BUILD/python-jsonpickle-3.0.4-build/jsonpickle-3.0.4/jsonpickle/unpickler.py", line 289, in loadrepr
    return eval(evalstr)
  File "<string>", line 1, in <module>
NameError: name 'datetime' is not defined. Did you forget to import 'datetime'?
/builddir/build/BUILD/python-jsonpickle-3.0.4-build/jsonpickle-3.0.4/jsonpickle/unpickler.py:278: UnexpectedException
___________________ PicklingTestCase.test_thing_with_module ____________________

self = <jsonpickle_test.PicklingTestCase testMethod=test_thing_with_module>

    def test_thing_with_module(self):
        obj = Thing('with-module')
        obj.themodule = os
    
        flattened = self.pickler.flatten(obj)
>       inflated = self.unpickler.restore(flattened)

tests/jsonpickle_test.py:392: 
[...]

PR #505, opened 2024-06-01
fix: pass mylocals to eval() for Python 3.13

davvid:
Newer versions of python need mylocals to be specified in loadrepr().

Fixes: #504


davvid:
@musicinmybrain confirmed in #504 that this fixed the tests for python3.13 in his environment so I'm going to go ahead and merge this now.

I'll look into tagging and releasing the next version soon.

PR #507, opened 2024-06-08
Add support for (de)serializing NoneType correctly

bhaveshpandey:
The present `loadclass` implementation fails to correctly loadclass for `builtins.NoneType`.

The PR adds support for this such that decoding will behave appropriately for following example:

```py

some_parms = dict( value_type=(str, type(None)), default=None, some_other_key=42 )
some_parms_enc = jsonpickle.encode(some_parms)
some_parms_dec = jsonpickle.decode(some_parms_enc)

some_parms == some_pars_dec
```

### DIFF
--- a/jsonpickle/unpickler.py
+++ b/jsonpickle/unpickler.py
@@ -183,6 +183,9 @@ def loadclass(module_and_name, classes=None):
     0
 
     """
+    # NoneType is a special case and can not be imported/created
+    if module_and_name == "builtins.NoneType":
+        return type(None)
     # Check if the class exists in a caller-provided scope
     if classes:
         try:
--- a/jsonpickle/unpickler.py
+++ b/jsonpickle/unpickler.py
@@ -183,9 +183,6 @@ def loadclass(module_and_name, classes=None):
     0
 
     """
-    # NoneType is a special case and can not be imported/created
-    if module_and_name == "builtins.NoneType":
-        return type(None)
     # Check if the class exists in a caller-provided scope
     if classes:
         try:
@@ -211,6 +208,9 @@ def loadclass(module_and_name, classes=None):
             return obj
         except (AttributeError, ImportError, ValueError):
             continue
+    # NoneType is a special case and can not be imported/created
+    if module_and_name == "builtins.NoneType":
+        return type(None)
     return None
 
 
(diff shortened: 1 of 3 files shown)


tests/jsonpickle_test.py
```diff
@@ -257,6 +257,12 @@ def test_list(self):
         self.assertEqual(listD, self.pickler.flatten(listD))
         self.assertEqual(listD, self.unpickler.restore(listD))
 
+    def test_NoneType(self):
```
Theelx, line 260: Can you name this ``test_nonetype``? Mixing snake_case and PascalCase, while it's accurate to both the stdlib name and jsonpickle's general formatting, is a bit jarring to read. Apart from that, thanks so much for writing a test! Most first-time contributors don't do that, I forgot when it was my first PR!

Theelx (changes requested):
Looks great to me, apart from a few relatively unimportant nitpicks. Only other thing is a CHANGELOG entry would be nice, but I can do that if you'd rather me take care of it.

jsonpickle/unpickler.py
```diff
@@ -183,6 +183,9 @@ def loadclass(module_and_name, classes=None):
     0
 
     """
+    # NoneType is a special case and can not be imported/created
+    if module_and_name == "builtins.NoneType":
```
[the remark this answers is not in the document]
bhaveshpandey: Makes sense, this should be addressed in a new push to this PR

tests/jsonpickle_test.py
bhaveshpandey: That's reasonable. Addressed this in an update to the PR. Thanks for a quick review!

bhaveshpandey:
> Looks great to me, apart from a few relatively unimportant nitpicks. Only other thing is a CHANGELOG entry would be nice, but I can do that if you'd rather me take care of it.

It would be great if you could do it for me please @Theelx ? I am not sure how I should go about adding to the changelist.

davvid (changes requested):
Please rebase+squash these three commits into a single commit and I  think this is okay to merge. 

<strike>In the meantime I'm going to go ahead and release v3.1.0 from what's currently in git as-is.

Once we merge this we can include this in the smaller v3.1.1 update unless anyone thinks otherwise that we should include this in v3.1.0.</strike>

UPDATE: I tagged v3.1.0 (and v3.2.0) so this'll be in the upcoming v3.2.1 release once it's merged.

Theelx (approved):
Looks good to me, I'll go ahead and add the CHANGELOG entry, then I can squash and merge for davvid!

bhaveshpandey:
Apologies for a delayed response on this, got caught up with work.

Thanks a lot for reviewing the PR and merging this on such a short notice @Theelx , @davvid .

PR #508, opened 2024-06-09
unpickler: avoid creating and returning temporary functions

davvid:
Add a private _passthrough() function that we can return whereever
we would have created a temporary function. This avoids creating
temporary functions and the associated extra allocations that
come along with them.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,7 @@
 Upcoming
 ========
     * ``tests/benchmark.py`` was updated to avoid python2 syntax.
+    * The unpickler was updated to avoid creating temporary functions.
 
 v3.2.1
 ======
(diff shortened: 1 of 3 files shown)


Theelx (commented):
When I ran benchmarks, there seemed to be a 5-10% speedup for simpler cases (string keys + no containers like lists/sets/dicts) for unpickling functions, while others stayed the same, and more complex ones with non-string keys and containers seemed to slow down. I'm going to do a more rigorous analysis of the results though, as this is currently based on eyeballing it. I would appreciate if we can wait to merge this until I can do statistical tests.

Theelx:
Ok, so this modification is ~0.35% faster than the current main across all the benchmarking tests on Python 3.9. However, the results show that it's 0.55% faster for encoding functions and 0.16% faster for decoding, which makes no sense as this doesn't touch encoding. I think it might be because of noise on my system, which would imply that this isn't a statistically significant change. I don't care whether this is merged, as this doesn't seem to positively or negatively impact speed, nor readability in my opinion. However, if you think it's easier to read, then feel free to merge it.

PR #509, opened 2024-06-09
SECURITY: add a security disclosure policy

davvid:


davvid:
Heads-up @Theelx  ~ Tidelift notified us asking that we add a security policy. They recommend letting them handle notification and disclosure so I've linked to their process here.

Theelx (approved):
LGTM!

PR #510, opened 2024-06-09
drop python<=3.7 support

kloczek:
According to https://endoflife.date/python python 3.7 has been EOSed 27 Jun 2023.
Filter all code over `pyupgrade --py38-plus`.

### DIFF
--- a/tests/object_test.py
+++ b/tests/object_test.py
@@ -905,13 +905,6 @@ def test_counter_roundtrip_with_keys():
     assert decoded.get(1) == 2
 
 
-issue281 = pytest.mark.xfail(
-    'sys.version_info >= (3, 8)',
-    reason='https://github.com/jsonpickle/jsonpickle/issues/281',
-)
-
-
-@issue281
 def test_list_with_fd():
     """Serialize a list with an file descriptor"""
     with open(__file__) as fd:
@@ -927,7 +920,6 @@ def _test_list_with_fd(fd):
     assert [None] == newobj
 
 
-@issue281
 def test_thing_with_fd():
     """Serialize an object with a file descriptor"""
     with open(__file__) as fd:
@@ -943,7 +935,6 @@ def _test_thing_with_fd(fd):
     assert newobj.name is None
 
 
-@issue281
 def test_dict_with_fd():
     """Serialize a dict with a file descriptor"""
     with open(__file__) as fd:
--- a/tests/object_test.py
+++ b/tests/object_test.py
@@ -14,7 +14,7 @@ from jsonpickle import compat, handlers, tags, util
 from jsonpickle.compat import queue
 
 
-class Thing(object):
+class Thing:
     def __init__(self, name):
         self.name = name
         self.child = None
@@ -51,7 +51,7 @@ class GetstateDict(dict):
         self.active = True
 
 
-class GetstateOnly(object):
+class GetstateOnly:
     def __init__(self, a=1, b=2):
         self.a = a
         self.b = b
@@ -60,7 +60,7 @@ class GetstateOnly(object):
         return [self.a, self.b]
 
 
-class GetstateReturnsList(object):
+class GetstateReturnsList:
     def __init__(self, x, y):
         self.x = x
         self.y = y
@@ -72,7 +72,7 @@ class GetstateReturnsList(object):
         self.x, self.y = state[0], state[1]
 
 
-class GetstateRecursesInfintely(object):
+class GetstateRecursesInfintely:
     def __getstate__(self):
         return GetstateRecursesInfintely()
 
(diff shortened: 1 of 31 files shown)


Theelx:
This will have to wait for jsonpickle 4.0.0 to release, which could be a few months. Dropping python compatibility is not something for a relatively minor library to do just because something is EOL, especially because we still get [50k downloads per day with 3.7](https://pypistats.org/packages/jsonpickle). I think we should wait for at least 3.8 to become EOL and for usage to drop, and then we can consider dropping 3.7 and 3.8 support.

davvid:
100% ~ since it doesn't burden jsonpickle much I've always tried to cast as wide a compatibility net as possible, even if it's a little extra code for us to maintain.

The stats and my personal experience both confirm that 3.7 is still widely used in some places, especially with third-party software. Take vfxplatform.com as an example.

It was the CY2021 specification that brought things up to 3.7. That means that vendors at the end of calendar year 2021 could release beta versions while still adhering to the spec and defer their final production "gold" versions until Q1 or early Q2 2022.

It can then take users (e.g. smaller studios with less resources or larger studios with larger code bases) time to recompile, repackage and test their in-house tools and swap production over to them when new versions of third-party tools become available. That adds another time delay, easily pushing things towards 2023 when a production might first start using 3.7 most heavily. Film production takes some time (let's say, 2 years plus more time for post-release one-offs or other reuse of assets) so it's easy to see how these tools will still be used through 2024 and 2025.

While most folks will be running with a pinned-down software stack that won't be changing, it's common to patch dependencies in response to security vulnerabilities or to upgrade specific parts of the stack to enable a new capability in an older software environment.

That generally makes me take the stance that it's best to meet users where they are (on older platforms/environments) and not cut them off unless there's a strong reason to do so. I'd personally save dropping 3.7 support for early 2025 or Q4 2024, which is more or less vfxplatform + 4 years. Maybe I'm a little extra cautious, but it doesn't cost jsonpickle too much IMO.

If you don't mind we can keep this PR open for a while and we can rebase it later once it's closer to showtime.

kloczek:
If in the setup.py will be `python_requires = ">=3.8"` pypi automatically will provide prev version.

Theelx:
> If in the setup.py will be `python_requires = ">=3.8"` pypi automatically will provide prev version.

That's true, but we still want to provide newer versions for people who are using 3.7. For example, the recently-merged PR to allow an alternative to the only remaining ``eval()`` path in ``loadrepr`` is beneficial to security, and dropping 3.7 support before that's released would mean that users of 3.7 don't get the benefits of being able to use 3.7. Additionally, dropping support on our API surface for a previously supported language version like CPython 3.7 without bumping the major version number would be a violation of semantic versioning. While many projects violate SemVer sometimes (usually on accident), I think it's better to avoid doing it on purpose so that we can increase trust of users in the stability of our package.

davvid:
@kloczek I just rebased and re-authored your changes against the latest main branch. We're planning on merging this soon.

Heads-up @Theelx

kloczek:
Thank you 👍

tests/object_test.py
```diff
@@ -936,7 +936,7 @@ def test_thing_with_fd():
 @issue281
 def test_dict_with_fd():
     """Serialize a dict with a file descriptor"""
-    fd = open(__file__, 'r')
+    fd = open(__file__)
```
Theelx, line 939: I think we should just change this and other instances of ``fd = open(__file__); fd.close()`` to the ``with open(__file__) as fd:`` context manager, as that's recommended nowadays.

Theelx (approved):
I'd suggest changing the open() thing to a context manager since pyupgrade doesn't seem to know about that recommendation, but apart from that it looks ready to merge!

tests/object_test.py
davvid: Good catch. I'll add a commit to get these using context managers and then start the merges shortly.

davvid: One small semantic change is that previously we were serializing a closed `fd` file descriptor (because we explicitly call `fd.close()` before using it) and now we'd be serializing an open file descriptor if we were to run the test code inside of the `with` block.

To keep the tests identical to their original behavior we can use `fd` after the `with` context block, but I've always considered that a slightly questionable pattern (with a paranoia that Python might decide to limit the scope of `with` context variable to just the context block one day).

Right now the `fd` variable outlives the block and it'll be closed, so the behavior should be identical. I wouldn't be surprised if some static analysis tool were to start warning about using `with` context-created variables after the block in which they were originally scoped, but that isn't the case now so I'm okay with using `fd` that way for now.

On the flip side, it's kinda interesting that all of our tests were testing a closed file descriptor. Things pass fine if we were to move the test code inside of the `with` block, so I'm kinda leaning towards doing that so that the code reads more simply and so that we actually exercise what happens when an open fd is run through jsonpickle.

Once we do that then we won't be testing closed file descriptors, so I'll add a separate test just for that scenario.

davvid: I ended up making all of the `open(...)` tests look more or less like this in `object_test.py`:

```python
def test_list_with_fd():
    """Serialize a list with an file descriptor"""
    with open(__file__) as fd:
        _test_list_with_fd(fd)
    _test_list_with_fd(fd)  # fd is closed.


def _test_list_with_fd(fd):
    """Serialize a list with an file descriptor"""
    obj = [fd]
    jsonstr = jsonpickle.encode(obj)
    newobj = jsonpickle.decode(jsonstr)
    assert [None] == newobj
```

I'm cool with that.

PR #511, opened 2024-06-09
Add basic benchmarking scripts

Theelx:
Currently, we've been relying on eyeballing the pytest-benchmark histograms generated between different changesets to catch performance regressions, which is prone to human error. This PR adds a bash script to run then analyze benchmarks, called ``benchmarking/analyze_benchmarks.sh``. At the moment, this allows us to see the overall speed change between the new changeset versus the main branch, plus the speed change for encoding and decoding functions. These scripts will be expanded eventually to allow for more types of filtering and visualizations, but they'll have to do for now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ __pycache__
 # Local benchmarking files for Theelx
 kern*.py
 *.lprof
+benchmark-*.json
+perf.csv
+perf.svg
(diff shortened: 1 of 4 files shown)


davvid:
Good stuff, looks good to me. Thanks again @Theelx !

davvid (approved):

PR #513, opened 2024-06-20
unpickler: avoid use of eval() in loadrepr

davvid:
We can eliminate the use of eval for the use case of loading modules by parsing the dotted-name lookup ourselves.

### DIFF
--- a/jsonpickle/unpickler.py
+++ b/jsonpickle/unpickler.py
@@ -278,6 +278,10 @@ def loadrepr(reprstr):
     """Returns an instance of the object from the object's repr() string.
     It involves the dynamic specification of code.
 
+    ..warning::
+
+        This function is unsafe and uses `eval()`. `loadrepr` is deprecated.
+
     >>> obj = loadrepr('datetime/datetime.datetime.now()')
     >>> obj.__class__.__name__
     'datetime'
@@ -290,6 +294,25 @@ def loadrepr(reprstr):
         localname = module.split('.', 1)[0]
     mylocals[localname] = __import__(module)
     return eval(evalstr, mylocals)
+
+
+def loadmodule(module_str):
+    """Returns a reference to a module.
+
+    >>> fn = loadmodule('datetime/datetime.datetime.fromtimestamp')
+    >>> fn.__name__
+    'fromtimestamp'
+
+    """
+    module, identifier = module_str.split('/')
+    toplevel_name = module.split('.', 1)[0]
+    result = __import__(module)
+    for name in identifier.split('.')[1:]:
+        try:
+            result = getattr(result, name)
+        except AttributeError:
+            return None
+    return result
 
 
 def has_tag_dict(obj, tag):
@@ -555,7 +578,7 @@ class Unpickler(object):
         if self.safe:
             # eval() is not allowed in safe mode
             return None
-        obj = loadrepr(obj[tags.REPR])
+        obj = loadmodule(obj[tags.REPR])
         return self._mkref(obj)
 
     def _loadfactory(self, obj):
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+v3.3.0
+======
+    * The unpickler was updated to avoid using ``eval``, which helps improve its
+      security. (+513)
+
 v3.2.2
 ======
     * A bug with the incorrect (de)serialization of NoneType objects has been fixed.


jsonpickle/unpickler.py
```diff
@@ -292,6 +296,28 @@ def loadrepr(reprstr):
...
+
+    >>> obj = loadmodule('datetime/datetime.datetime')
+    >>> obj.__name__
+    'datetime'
+
+    """
+    module, evalstr = module_str.split('/')
+    mylocals = locals()
+    localname = module
+    if '.' in localname:
```
Theelx, line 310: I don't recall the exact semantics of loading a module, but is it possible for there to be 2+ dots, like if someone tries to import something like a function inside a class inside a file (for example, datetime/datetime.datetime.fromtimestamp)? This might not even be possible in python though, I'm unclear. If it is possible, we should add a test for that case.

jsonpickle/unpickler.py
```diff
@@ -555,7 +581,7 @@ def _restore_repr(self, obj):
         if self.safe:
             # eval() is not allowed in safe mode
             return None
-        obj = loadrepr(obj[tags.REPR])
+        obj = loadmodule(obj[tags.REPR])
```
Theelx, line 584: While what follows isn't a change for now, it would be nice eventually to rename tags.REPR to tags.MODULE or something like that, and provide a helper for decoding tags.REPR strings. This could be done before 4.0.0, as long as the helper function is invoked automatically when an instance of tags.REPR is detected. It could also smooth the transition to a new tags.MODULE if it saves written/updated using the new tags.MODULE.

jsonpickle/unpickler.py
```diff
@@ -292,6 +296,28 @@ def loadrepr(reprstr):
...
 
+def loadmodule(module_str):
+    """Returns a reference to a module.
+
+    >>> obj = loadmodule('datetime/datetime.datetime')
+    >>> obj.__name__
+    'datetime'
+
+    """
+    module, evalstr = module_str.split('/')
```
Theelx, line 307: Even though importing arbitrary files might use an eval in it somewhere inside Python, it would probably be best not to name it evalstr if we aren't explicitly eval-ing, to avoid potential confusion from anyone reading the code.

davvid:
> Does calling locals() have some sort of effect on the function's state, like globals()?

Good catch! `locals()` does let you manipulate the variables local to the function, but in this case I actually forgot to delete that code and it's an unused variable.

> is it possible for there to be 2+ dots, like if someone tries to import something like a function inside a class inside a file (for example, datetime/datetime.datetime.fromtimestamp)? If it is possible, we should add a test for that case.

Yes, imports with multiple dots is supported. It's a little subtle, but I just tweaked the function to be a little clearer, at least naming-wise. The part that handles the dotted lookup is this loop, which loops over the dotted names and progressively calls `getattr(...)` to dig into the nested `.` object.


```python
    for name in identifier.split('.')[1:]:
        try:
            result = getattr(result, name)
```


> it would be nice eventually to rename tags.REPR to tags.MODULE or something like that, and provide a helper for decoding tags.REPR strings.

That's a great idea. I've started experimenting with that in #514, which uses this branch as its base.


> Even though importing arbitrary files might use an eval in it somewhere inside Python, it would probably be best not to name it evalstr if we aren't explicitly eval-ing, to avoid potential confusion from anyone reading the code.

Definitely. I renamed the variables to be clearer in this latest version.

I think this addresses all these points, so I'll let this cook until Sunday before merging. Thanks for the thorough review.

davvid:
One notable change in this latest is that I removed the deprecation notice that was originally added to the release notes. The changes from #514 suggest that maybe we don't need to deprecate `loadrepr` but rather just make it disabled by default since we might want to keep it around for data restoration purposes.

It's there to support reading jsonpickle files from older versions, so there's the backwards-compat reason to keep loadrepr around for as long as we have the `safe=False` code path.

In that case.. since it's unclear, marking this function deprecated seemed premature so I removed that part.

I also wonder.. should we call this new function `_loadmodule` with a leading underscore to mark it as "don't use this, it's internal"?

We can't use two underscores because python does something unexpected when trying to call a free-standing function with a leading dunder from within a method inside a class -- python isn't able to call the free-standing function because it name-mangles it to contain the class name.

This almost seems like a bug.. but I guess this is normal?

```python
def __ok():
    pass

class Example:
    def __init__(self):
        __ok()

example = Example()
```
Running this emits:

```
  File "scratch.py", in main
    example = Example()
              ^^^^^^^^^
  File "scratch.py", in __init__
    __ok()
    ^^^^
NameError: name '_Example__ok' is not defined
```

Maybe a TODO would be to document that functions named with a leading underscore are not considered part of the public API.

I'm tempted to rename `loadmodule` to `_load_module` to keep the API surface smaller by default, but if you think this might be a useful thing to keep for use by custom handlers then I'm fine with keeping it as-is like we have here and adding it to the public API.

Theelx:
I think you're right that it would be premature to deprecate the loadrepr function, especially because it could still be useful for loading files generated by older versions of jsonpickle. Regarding whether `loadmodule` should be marked with a leading underscore, it would be nice to mark it as internal to reduce the API surface (as you mention) but the fact that loadrepr isn't internal and that `loadmodule` would be replacing `loadrepr` indicates to me that there shouldn't be a leading underscore. That way, anyone who may be using `loadrepr` in their own scripts (which has been fine up until this point) can migrate to the replacement easily.

The point you bring up about name-mangling is interesting. It's actually already worked around in a different context within jsonpickle [here](https://github.com/jsonpickle/jsonpickle/blob/66a812f63912d0352b6d79f24a6c59f3e94aa6e2/jsonpickle/pickler.py#L477)., but if `loadmodule` is not private anyway then I guess it doesn't really matter.

Theelx (approved):

davvid:
> the fact that loadrepr isn't internal and that loadmodule would be replacing loadrepr indicates to me that there shouldn't be a leading underscore

:100:

PR #514, opened 2024-06-29
tags: use py/mod tags for modules instead of py/repr

davvid:
A continuation of the ideas from #513 which we can refine after that PR is merged.

### DIFF
--- a/jsonpickle/unpickler.py
+++ b/jsonpickle/unpickler.py
@@ -19,7 +19,7 @@ def decode(
     context=None,
     keys=False,
     reset=True,
-    safe=False,
+    safe=True,
     classes=None,
     v1_decode=False,
     on_missing='ignore',
@@ -45,8 +45,13 @@ def decode(
         This flag is not typically used outside of a custom handler or
         `__getstate__` implementation.
 
-    :param safe: If set to True, eval() is avoided, but backwards-compatible
-        (pre-0.7.0) deserialization of repr-serialized objects is disabled.
+    :param safe: If set to False, use of ``eval()`` for backwards-compatible (pre-0.7.0)
+        deserialization of repr-serialized objects is enabled. Defaults to True.
+
+        .. warning::
+
+            ``eval()`` is used when set to ``False`` and is not secure against
+            malicious inputs. You should avoid setting ``safe=False``.
 
     :param classes: If set to a single class, or a sequence (list, set, tuple) of
         classes, then the classes will be made available when constructing objects.
@@ -296,10 +301,10 @@ def loadrepr(reprstr):
     return eval(evalstr, mylocals)
 
 
-def loadmodule(module_str):
+def _loadmodule(module_str):
     """Returns a reference to a module.
 
-    >>> fn = loadmodule('datetime/datetime.datetime.fromtimestamp')
+    >>> fn = _loadmodule('datetime/datetime.datetime.fromtimestamp')
     >>> fn.__name__
     'fromtimestamp'
 
@@ -341,7 +346,7 @@ class Unpickler(object):
         self,
         backend=None,
         keys=False,
-        safe=False,
+        safe=True,
         v1_decode=False,
         on_missing='ignore',
         handle_readonly=False,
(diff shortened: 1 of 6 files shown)


tests/datetime_test.py
```diff
@@ -241,9 +241,10 @@ def test_datetime_repr_not_unpicklable(self):
         obj = datetime.datetime.now()
         pickler = jsonpickle.pickler.Pickler(unpicklable=False)
         flattened = pickler.flatten(obj)
-        self.assertFalse(tags.REPR in flattened)
-        self.assertFalse(tags.OBJECT in flattened)
-        self.assertEqual(obj.isoformat(), flattened)
+        assert tags.REPR not in flattened
```
Theelx, line 244: Is there a reason not to use the `self.assertBoolean` functions? I would think it'd be easier as it would allow us to easily modify behavior of the functions if needed for whatever reason someday, plus for consistency with the rest of the existing tests.

jsonpickle/unpickler.py
```diff
@@ -19,7 +19,7 @@ def decode(
     context=None,
     keys=False,
     reset=True,
-    safe=False,
+    safe=True,
```
Theelx, line 22: I have a gut feeling that this is going to come back to bite us in that loadmodule isn't compatible with some use case, but I can't find evidence for this feeling so YOLO I guess.

Theelx (approved):

tests/datetime_test.py
davvid: Yeah, there's some logic to this, though it will look inconsistent from the perspective of the individual diffs.

The older camelCased `assertBoolean` methods are from the old days when the `unittest` module was new and the names were chosen to match the exact names of the Java junit library, which exposed a similar API.

One of pytest's selling points is that it lets us use a consistent naming convention by by leveraging the language's existing `assert` statements. pytest augments `assert` statements by adding pretty printing and diffing, which makes the advantages of using the junit-inspired methods almost non-existent. It does drop the ability to extend these methods, but I hope we won't need to do that :crossed_fingers: .

So over time I've taken the approach of replacing older tests' assertBoolean methods with pytest-native equivalents as I run into them updating the test method for some other reason. I'll just do this incrementally as tests get updated, though at some point I might make a series of file-by-file commits to updates the test suite all at once and leverage more of pytest's niceties. Until then the diffs will look like they're inconsistent with their surroundings, but it's okay since it should be incrementally going towards the newer pytest style.

While I find that `asserts` are shorter, simpler, and easier to read/type, it's mostly aesthetic as both ultimately do the same thing. The python core team decided that they'll keep the junit names around forever so there's not any pressing deprecation or technical reason to change it on any timeframe.

jsonpickle/unpickler.py
davvid: Yeah, this is probably the part that is worth a major version bump just for extra safety.

The question I faced here was, "should jsonpickle be safe by default, or opt-in to safety?". The correct answer should really be "safe by default," but it meant changing semantics. It's hard to know if anyone was relying on it so the changelog will need to be updated to call this out more.

Unfortunately we won't find out if this affects anyone until we put it out there and wait 3-5 years for all the slow upgraders to come out of the woodwork. As long as we have an easy way (`safe=False`) for users to opt-in to the compatible behavior then we should end up in a better place as they'll be consciously choosing to use the unsafe mode.

tests/datetime_test.py
Theelx: Ah, that makes sense, thanks! I forgot that we use pytest, and from the POV of pytest it makes total sense to use asserts and not the unittest functions.

jsonpickle/unpickler.py
Theelx: I think you're right, jsonpickle should absolutely be safe by default. I do think that before this is merged, the CHANGES.rst file should be updated to specify that `safe=False` is an alternative if the user so chooses. I'll go ahead and update that now, and you can feel free to edit it and then merge it if you want.

davvid:
I squashed in the typofix you committed and also squashed in a further update to the changelog that calls out the `safe=...` change as a breaking change.

Since I think this might warrant a major version bump we can hold onto these changes here and defer merging them until after the next v3.x release. We should probably let #515 play out and once once that's merged we'll tag the next v3.x release. These changes can follow behind that merge as the starting point for v4.x.

davvid:
@Theelx I almost forgot about this one ;-)

I just rebased this against the latest and I think this is more or less about ready to go if you're still cool with the caveats we discussed in this PR.

PR #515, opened 2024-06-30
Issue 126: Allow objects with non-data encoding iterators to be handled.

dwagon:
If an object is an iterator and also has its own data then the data won't be encoded, just the iterator. This PR adds a flag to ignore the iterator.

A better solution would be to encode both iterator and data.

### DIFF
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -35,7 +35,7 @@ def encode(
     separators=None,
     include_properties=False,
     handle_readonly=False,
-        ignore_iterator=False
+    ignore_iterator=False
 ):
     """Return a JSON formatted representation of value, a Python object.
 
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -35,6 +35,7 @@ def encode(
     separators=None,
     include_properties=False,
     handle_readonly=False,
+        ignore_iterator=False
 ):
     """Return a JSON formatted representation of value, a Python object.
 
@@ -134,6 +135,10 @@ def encode(
         basically prevents jsonpickle from raising an exception for such objects.
         You MUST set ``handle_readonly=True`` for the decoding if you encode with
         this flag set to ``True``.
+    :param ignore_iterator:
+        Handle objects that contain an iterator that doesn't encode data from that object.
+        If set to False, the default, an object with an iterator will ignore other data elements.
+        Defaults to ``False``.
 
     >>> encode('my string') == '"my string"'
     True
@@ -161,6 +166,7 @@ def encode(
         include_properties=include_properties,
         handle_readonly=handle_readonly,
         original_object=value,
+        ignore_iterator=ignore_iterator
     )
     return backend.encode(
         context.flatten(value, reset=reset), indent=indent, separators=separators
@@ -210,6 +216,7 @@ class Pickler(object):
         include_properties=False,
         handle_readonly=False,
         original_object=None,
+        ignore_iterator=False
     ):
         self.unpicklable = unpicklable
         self.make_refs = make_refs
@@ -247,6 +254,7 @@ class Pickler(object):
         self.include_properties = include_properties
 
         self._original_object = original_object
+        self.ignore_iterator = ignore_iterator
 
     def _determine_sort_keys(self):
         for _, options in getattr(self.backend, '_encoder_options', {}).values():
@@ -681,7 +689,7 @@ class Pickler(object):
         if util.is_sequence_subclass(obj):
             return self._flatten_sequence_obj(obj, data)
 
-        if util.is_iterator(obj):
+        if not self.ignore_iterator and util.is_iterator(obj):
             # force list in python 3
             data[tags.ITERATOR] = list(map(self._flatten, islice(obj, self._max_iter)))
             return data
--- a/tests/object_test.py
+++ b/tests/object_test.py
@@ -188,6 +188,25 @@ class ThingWithDefaultFactory:
         # Keep track of how many times this constructor has been run.
         ThingWithDefaultFactoryRegistry.count += 1
 
+class ThingWithIteratorPlusData:
+    """ Class that has an iterator and more"""
+    def __init__(self, a):
+        self.a = a
+        self._data = [3,1,4,1,5,9]
+        self._index = 0
+
+    def __iter__(self):
+        self._index = 0
+        return self
+
+    def __next__(self):
+        try:
+            item = self._data[self._index]
+        except IndexError as e:
+            raise StopIteration from e
+        self._index += 1
+        return item
+
 
 class IntEnumTest(enum.IntEnum):
     X = 1
@@ -1139,6 +1158,13 @@ def test_datetime_with_tz_copies_refs():
     assert len(decoded) == 2
     assert decoded[0] == d0
     assert decoded[1] == d1
+
+def test_iterator_plus():
+    obj = ThingWithIteratorPlusData(1)
+    encoded = jsonpickle.encode(obj, ignore_iterator=True)
+    decoded = jsonpickle.decode(encoded)
+    assert hasattr(decoded, 'a')
+    assert decoded.a == obj.a
 
 
 if __name__ == '__main__':


tests/object_test.py
```diff
@@ -188,6 +188,25 @@ def __init__(self):
...
+
+    def __iter__(self):
+        self._index = 0
+        return self
+
+    def __next__(self):
+        try:
+            item = self._data[self._index]
+        except IndexError as e:
+            raise StopIteration from e
```
Theelx, line 206: Is the ``from e`` syntax compatible with Python versions as low as 3.7? I could have sworn that ``raise Exception from e`` was a newer construct, but I may be wrong.

jsonpickle/pickler.py
```diff
@@ -35,6 +35,7 @@ def encode(
     separators=None,
     include_properties=False,
     handle_readonly=False,
+        ignore_iterator=False
```
Theelx, line 38: This could be an artifact of the github web formatting, but it looks to be indented here at one level past the other options. Could you format the code with black just in case it isn't an artifact?

Theelx (approved):
Looks great to me, thanks for the PR!

tests/object_test.py
Theelx: It seems the tests passed on 3.7, so I guess this is all good.

Theelx:
(I'll add the CHANGES.rst entry after this is merged)

dwagon:
The more I think about this the more convinced I am that it isn't the correct approach. It is too broad a brush and will cause just as many problems as it solves. 

Either need to support iters and dicts in the same class; or be able to specify on a per class basis (through a ``__jsonpickle`` attribute?) which attributes to include / exclude.

The latter one seems to be the best (and probably easiest to implement) as there are occasionally members of a class that you might not want to save / restore (e.g. temporary data)

Theelx:
> The more I think about this the more convinced I am that it isn't the correct approach. It is too broad a brush and will cause just as many problems as it solves.

It's quite possible that that's correct, but we also can't let perfect be the enemy of good. If it's indeed the case that it causes more problems than it solves, then yeah it shouldn't be merged, but otherwise I think it would be useful to try it out.
> Either need to support iters and dicts in the same class; or be able to specify on a per class basis (through a __jsonpickle attribute?) which attributes to include / exclude.

If I'm understanding what you mean by a "``__jsonpickle`` attribute" (a dictionary attribute on the class + instance being pickled that maps attributes to booleans representing whether the attribute should be in/excluded), I don't think such an attribute would be the best immediate solution for that. A user can already customize the attributes that they want to be encoded/decoded by setting their own ``__getstate__()`` or ``__setstate__()`` magic methods on the class. I would definitely try to support both iters and dicts instead, as I don't believe there's a workaround for that. Maybe we could include such a ``__jsonpickle`` attribute later on, as I can see it being easier for beginners to make use of, especially since magic methods aren't well documented currently.

jsonpickle/pickler.py
Theelx: Fixed it myself. Waiting to merge now until discussion on the PR is resolved to a satisfactory extent.

davvid (approved):

davvid:
> Maybe we could include such a __jsonpickle attribute later on, as I can see it being easier for beginners to make use of, especially since magic methods aren't well documented currently.

Agreed. We might want to call it e.g. `_jsonpickle_fields` to avoid the double-underscore name mangling and to leave space for future enhancement, but that's all stuff we can figure out later.

I'm very curious to see if there's code out there that would fail with `ignore_iterator=True`.

That said... I'm starting to slightly question this change slightly. Semantically, I would imagine that ignoring iterators would mean that we might encode them to `None`, but what we actually do is make it so that we include their attributes.

That makes me wonder whether we should name this parameter to better prepare the naming for future enhancement where we improve the iterator+data code path.

It's interesting that we currently blindly consume iterators and encode its data as a list, so it does seem like a future goal would be to have encoding of custom iterator data become the new default, and ideally even eliminating the emission of tags.ITERATOR if we can handle iterators and regular objects uniformly.

The code currently has two aspects that we want the name to reflect ~ the iter-to-list conversion and the custom iterator class + data aspect. Maybe a clearer name might be something like, `include_iterator_attributes`?

That name is also similar to the `include_properties` option we currently support, which is kinda nice as well.

This is just cosmetic though, so I approved this but we can still change this before (or even after) merging as long as we do it before tagging the next release.

Theelx:
> Semantically, I would imagine that ignoring iterators would mean that we might encode them to None, but what we actually do is make it so that we include their attributes.

That's interesting, as I'd expect that ignoring iterators would mean excluding them from the result entirely. Maybe we should bikeshed the name more, as you're suggesting. I think ``include_iterator_attributes`` is accurate but a bit verbose for a parameter though, so maybe it might be better to call it ``include_iterator_attrs``? I don't know if that'd be less clear to non-native English speakers though, I'll need to check with friends of mine who don't speak English natively first.

Overall, I'd wait on merging until dwagon has a chance to respond to these ideas.

dwagon:
I'm curious as to the reasoning / use case to encode iterators at all. 
My use of iterators fall in to two major categories:

- Iterating across a container inside a class - no need for saving the iterator; the container should be saved
- Iterating across an external data source (file, database, etc) - no need to save the iterator; just save the data source

Is there something that I have missed or a technique I should have been using?

The attribute should be either to explicitly include or exclude members of a class (and by default do everything). It shouldn't try to do both (having an "include" list and an "exclude" list is just going to cause confusion). As the default is to pickle everything, I'm thinking specifying attributes to not include is more logical. It means that if a user adds a new attribute to a class they are trying to pickle it will work with no additional effort - only if there is a problem would they need to adjust the field list. 

As to the naming of the attribute:

- It should include the name "jsonpickle" to avoid clashes with anything else.
- It should be terse; we're not coding cobol :)

My suggestion would be "_jsonpickle_exclude"

dwagon:
One reason I dislike the "ignore_iterators" flag is that it is coarse grain - it affects all of the saved classes whether you want it to or not. So if you have a class that you want to have iters and one where you don't you would have to treat them separately - and if they are nested you are out of luck. Whereas a attribute on a class means you have fine grained control when you need it.

dwagon:
I created a PR (https://github.com/jsonpickle/jsonpickle/pull/517) with code that does the exclusion.
It works but there are probably huge gaps in cases; and almost certainly easier/better ways of doing it.

If you like the concept I will do the docs as well.

davvid:
> include_iterator_attrs? I don't know if that'd be less clear to non-native English speakers though, I'll need to check with friends of mine who don't speak English natively first.

I'd :+1: for the shorter name and if we go down that road. I'm super curious to know what your friends think of the attrs abbreviation. To me, one attr in favor of attr is that python already uses the abbreviation in getattr, setattr, hasattr and friends.

Better still, it seems like that might be a moot point if we can do without it altogether. I like the concept and diffs in #517 better, especially since it's minimal, general-purpose, and fine-grained.

> One reason I dislike the "ignore_iterators" flag is that it is coarse grain

Same.

> (#517) works but there are probably huge gaps in cases; and almost certainly easier/better ways of doing it.
> 
> If you like the concept I will do the docs as well.

Docs would be much appreciated. There may be gaps but since it's something that only affects new code that opts-in then it should be pretty easy to spackle over the gaps as users encounter them. I'm optimistic that we can make it work.

What do you think @Theelx? I'll start reviewing #517 in the meantime. Thanks for taking the time to get this going.

Theelx:
> I'm super curious to know what your friends think of the attrs abbreviation.

I asked 3 friends, one who speaks French, one who speaks Spanish, and one who speaks Korean (all natively). They all understand the attrs abbreviation, though I forgot to ask them whether they like attrs or attributes better.

> One reason I dislike the "ignore_iterators" flag is that it is coarse grain

Same, additionally I think we have so many options on encode and decode that it'd be worth thinking about a redesign of some of them for 4.0.0, specifically to merge some functionality together. I emailed davvid about this a while back, but I really liked a bitflags/enum approach to specifying flags, and enums are already in the standard library.

> I'm optimistic that we can make it work.

Absolutely! Thank you so much for your time, help, and well-reasoned approaches @dwagon, I think these contributions are going to be a big benefit to jsonpickle overall!

davvid:
> I think we have so many options on encode and decode that it'd be worth thinking about a redesign of some of them for 4.0.0, specifically to merge some functionality together. I emailed davvid about this a while back, but I really liked a bitflags/enum approach to specifying flags, and enums are already in the standard library.

I'm on board with this idea as a reason to bump major versions once we get around to it.

We could keep the existing names for backwards-compatibility forever and introduce new functions with the better design, but I suspect we'll want to drop the old baggage eventually to prevent the implementation from metastasizing.

Either way, we'd probably want to implement it initially by first having the existing API call into a redesigned API that uses bitflags/enum. We can do that before bumping major versions, which is nice since it'll exercise the new API by ensuring that we can reflect our current use cases before swapping over.

We'd then have a commit that changes the signature of the existing functions to remove the old parameter names and that'd be major version trigger.

Anyways, we're off on a tangent so I'll try and keep focused by closing this PR for now since I think we're all in alignment here. Thanks for sharing and helping shape things.

PR #516, opened 2024-07-07
Add Usage section to README

Theelx:
This addresses part of #484 by making a short example script which shows how jsonpickle can be used. Todo would be to include an even shorter code snippet in the README itself, as with this change the README just links to the newly-made examples directory in the GitHub repo. This PR also adds some basic utilities for future example scripts, which could showcase numpy extension support, getstate usage, and more advanced topics.

### DIFF
--- a/examples/save_class_to_file.py
+++ b/examples/save_class_to_file.py
@@ -1,0 +1,51 @@
+import jsonpickle
+
+from utilities import ensure_no_files_overwritten
+
+ensure_no_files_overwritten(
+    expected_contents='{"py/object": "__main__.Example", "data": {"BAR": 1, "foo": 0}}'
+)
+
+
+class Example:
+    def __init__(self):
+        self.data = {"foo": 0, "BAR": 1}
+
+    def get_foo(self):
+        return self.data["foo"]
+
+    def __eq__(self, other):
+        # ensure that jsonpickle preserves important stuff like data across encode/decoding
+        return self.data == other.data and self.get_foo() == other.get_foo()
+
+
+# instantiate the class
+ex = Example()
+# encode the class. this returns a string, like json.dumps
+encoded_instance = jsonpickle.encode(ex)
+assert (
+    encoded_instance
+    == '{"py/object": "__main__.Example", "data": {"BAR": 1, "foo": 0}}'
+)
+print(
+    f"jsonpickle successfully encoded the instance of the Example class! It looks like: {encoded_instance}"
+)
+
+
+with open("example.txt", "w+") as f:
+    f.write(encoded_instance)
+    print(
+        f"jsonpickle successfully wrote the instance of the Example class to example.txt!"
+    )
+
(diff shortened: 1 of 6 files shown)


examples/save_class_to_file.py
```diff
@@ -0,0 +1,51 @@
...
+assert (
+    encoded_instance
+    == '{"py/object": "__main__.Example", "data": {"BAR": 1, "foo": 0}}'
+)
+print(
+    f"jsonpickle successfully encoded the instance of the Example class! It looks like: {encoded_instance}"
+)
+
+
+with open("example.txt", "w+") as f:
```
davvid, line 35: Should we use `example.json` here and below instead of `.txt`?

examples/save_class_to_file.py
```diff
@@ -0,0 +1,51 @@
...
+    f.write(encoded_instance)
+    print(
+        f"jsonpickle successfully wrote the instance of the Example class to example.txt!"
```
davvid, line 38: We might be able to use a regular string rather than an f-string here.

davvid (approved):
Good stuff!

examples/utilities.py
```diff
@@ -0,0 +1,15 @@
+import os
+import sys
+
+
+def ensure_no_files_overwritten(expected_contents):
+    # ensure we don't overwrite the user's files, if they run it on their machine
+    if os.path.exists(os.path.join(os.getcwd(), 'example.txt')):
```
davvid, line 7: We call `open(...)` with just the basename below so we should be able to do the same here and drop the os.path.join parts.

examples/utilities.py
```diff
@@ -0,0 +1,15 @@
...
+        with open("example.txt", "r") as f:
+            contents = f.read()
+        if contents != expected_contents:
+            print(
+                "I don't want to overwrite the example.txt file in my current directory! Please delete example.txt and try again!"
+            )
+            # exit with a non-zero error code
+            sys.exit(1)
```
davvid, line 15: A neat trick with `sys.exit` is that you can pass it an error message string and it'll print it out and exit with a non-zero exit code. That lets you combine the `print` and `sys.exit` steps into a single call. It returns exit code 1 as well.

examples/save_class_to_file.py
Theelx: Yes, good point.

examples/save_class_to_file.py
Theelx: Opps yeah, I think I previously had something printed out but forgot to remove the f

examples/utilities.py
Theelx: Oh, I didn't know that, thanks!

examples/utilities.py
Theelx: Yeah fair, I just didn't want to worry about if Windows would throw a fit at doing that. I'll fix that now.

PR #517, opened 2024-07-13
Define _jsonpickle_exclude to omit attributes from encoding

dwagon:
Initial PR to talk about concept.

### DIFF
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -539,6 +539,7 @@ class Pickler(object):
         has_getnewargs_ex = util.has_method(obj, '__getnewargs_ex__')
         has_getinitargs = util.has_method(obj, '__getinitargs__')
         has_reduce, has_reduce_ex = util.has_reduce(obj)
+        exclude = set(getattr(obj, '_jsonpickle_exclude', ()))
 
         # Support objects with __getstate__(); this ensures that
         # both __setstate__() and __getstate__() are implemented
@@ -675,7 +676,7 @@ class Pickler(object):
             return data
 
         if util.is_dictionary_subclass(obj):
-            self._flatten_dict_obj(obj, data)
+            self._flatten_dict_obj(obj, data, exclude=exclude)
             return data
 
         if util.is_sequence_subclass(obj):
@@ -693,7 +694,7 @@ class Pickler(object):
 
             # hack for zope persistent objects; this unghostifies the object
             getattr(obj, '_', None)
-            return self._flatten_dict_obj(obj.__dict__, data)
+            return self._flatten_dict_obj(obj.__dict__, data, exclude=exclude)
 
         if has_slots:
             return self._flatten_newstyle_with_slots(obj, data)
@@ -770,7 +771,7 @@ class Pickler(object):
         data[k] = self._flatten(v)
         return data
 
-    def _flatten_dict_obj(self, obj, data=None):
+    def _flatten_dict_obj(self, obj, data=None, exclude=()):
         """Recursively call flatten() and return json-friendly dict"""
         if data is None:
             data = obj.__class__()
@@ -780,17 +781,17 @@ class Pickler(object):
         if self.keys:
             # Phase 1: serialize regular objects, ignore fancy keys.
             flatten = self._flatten_string_key_value_pair
-            for k, v in util.items(obj):
+            for k, v in util.items(obj, exclude=exclude):
                 flatten(k, v, data)
 
             # Phase 2: serialize non-string keys.
             flatten = self._flatten_non_string_key_value_pair
-            for k, v in util.items(obj):
+            for k, v in util.items(obj, exclude=exclude):
                 flatten(k, v, data)
         else:
             # If we have string keys only then we only need a single pass.
             flatten = self._flatten_key_value_pair
-            for k, v in util.items(obj):
+            for k, v in util.items(obj, exclude=exclude):
                 flatten(k, v, data)
 
         # the collections.defaultdict protocol
@@ -817,7 +818,7 @@ class Pickler(object):
         if hasattr(obj, '__dict__') and self.unpicklable and obj != obj.__dict__:
             if self._mkref(obj.__dict__):
                 dict_data = {}
-                self._flatten_dict_obj(obj.__dict__, dict_data)
+                self._flatten_dict_obj(obj.__dict__, dict_data, exclude=exclude)
                 data['__dict__'] = dict_data
             else:
                 data['__dict__'] = self._getref(obj.__dict__)
--- a/jsonpickle/util.py
+++ b/jsonpickle/util.py
@@ -577,9 +577,11 @@ def itemgetter(obj, getter=operator.itemgetter(0)):
     return compat.ustr(getter(obj))
 
 
-def items(obj):
+def items(obj, exclude=()):
     """
     TODO: Replace all calls to this with plain dict.items()
     """
     for k, v in obj.items():
+        if k in exclude:
+            continue
         yield k, v
--- a/tests/object_test.py
+++ b/tests/object_test.py
@@ -289,6 +289,20 @@ class ExceptionWithArguments(Exception):
         self.value = value
 
 
+class ThingWithExclusion:
+    _jsonpickle_exclude = ["foo"]
+
+    def __init__(self, a):
+        self.foo = 1
+        self.bar = a
+
+
+class ThingWithExcludeSubclass:
+    def __init__(self, foo):
+        self.foo = foo
+        self.thing = ThingWithExclusion(3)
+
+
 class AdvancedObjectsTestCase(SkippableTest):
     def setUp(self):
         self.pickler = jsonpickle.pickler.Pickler()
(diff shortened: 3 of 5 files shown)


jsonpickle/pickler.py
```diff
@@ -234,6 +234,7 @@ def __init__(
         self._flattened = {}
         # Used for util.is_readonly, see +483
         self.handle_readonly = handle_readonly
+        self.excludes = []
```
davvid, line 237: `self.excludes = set()`

jsonpickle/pickler.py
```diff
@@ -539,6 +540,7 @@ def _flatten_obj_instance(self, obj):
         has_getnewargs_ex = util.has_method(obj, '__getnewargs_ex__')
         has_getinitargs = util.has_method(obj, '__getinitargs__')
         has_reduce, has_reduce_ex = util.has_reduce(obj)
+        self.excludes = getattr(obj, '_jsonpickle_exclude') if hasattr(obj, '_jsonpickle_exclude') else []
```
davvid, line 543: `self.excludes = set(getattr(obj, '_jsonpickle_exclude', ()))` lets us do without the if-else expression.

It also lets us use a `set` object internally so it has a chance of being faster than `list` membership checks.

jsonpickle/pickler.py
```diff
@@ -780,17 +782,17 @@ def _flatten_dict_obj(self, obj, data=None):
         if self.keys:
             # Phase 1: serialize regular objects, ignore fancy keys.
             flatten = self._flatten_string_key_value_pair
-            for k, v in util.items(obj):
+            for k, v in util.items(obj, self.excludes):
```
davvid, line 785: `for k, v in utils.items(obj, exclude=self.excludes):`  --  include the kwarg name at the call site here and below since it's optional.

jsonpickle/util.py
```diff
@@ -577,9 +577,11 @@ def itemgetter(obj, getter=operator.itemgetter(0)):
     return compat.ustr(getter(obj))
 
 
-def items(obj):
+def items(obj, exclude_keys=[]):
```
davvid, line 580: Let's make this `def items(obj, exclude=()):` -- Python trivia/best-practice is to avoid mutable contains as defaults for function keyword parameters.

The rationale for this family of suggestions is that keyword parameters are stored as global variables as part of the `def` object, and code that inadvertently mutates them ends up mutating a global variable that accumulates state. We don't mutate stuff here, but we can stick to the spirit of avoiding that construct by defaulting to an empty tuple instead of a list.

This is not an issue in practice in this code since we don't mutate, but if you use lists, sets or dicts as default values it can cause experienced python users to do a double-take while reading/reviewing the code.

It's funny.. `util.items()` was originally added for python2+3 compatibility and has a TODO that we should get rid of it. If we're going to start relying on it then we should replace the docstring below to not suggest its removal.

tests/object_test.py
```diff
@@ -288,6 +288,17 @@ def __init__(self, value):
         super(ExceptionWithArguments, self).__init__('test')
         self.value = value
 
+class ThingWithExclusion:
+    _jsonpickle_exclude = ["foo"]
```
davvid, line 292: Apologies for the true boring style bikeshed, but please use single quotes like the rest of this file. I'll try and setup `cercis` for formatting the code soon since it enforces my extreme OCD better than `black` does.

You can run `garden fmt` (see `garden.yaml` for details) to deal with other minor stylistic issues if you already have `black` installed.

davvid (changes requested):
These are really minor sugs so once the doc updates are in place this looks good to me both conceptually and implementation-wise.

Please squash the suggested updates into this commit and let's see whether @Theelx concurs that we go further with this idea instead of the approach in #515.

jsonpickle/pickler.py
Theelx: I support this. Shouldn't it be able to be specified at creation time too though, like through the encode interface?

jsonpickle/pickler.py
Theelx: Why not just convert it to a set at instantiation time? ``_flatten_obj_instance`` is called very often, so I can see this being a big perf penalty. Alternatively, if we really care about performance, it could be a tuple for small tuples (<5 items), and a set for larger ones. If it's only being used in ``util.items()``, then the ``in`` check will work both ways, though if this route is taken then ``self.exclude`` should become ``self._exclude`` so the user doesn't mess with it after it's gone through jsonpickle's type coercion at instantiation. I'll try to have this measured more concretely by Wednesday though, I'd do it today but I have a midterm due at midnight today for a summer class :(

jsonpickle/pickler.py
Theelx: Agree.

jsonpickle/util.py
Theelx: Agree.

tests/object_test.py
Theelx: I commented this on the PR implementing cercis, but I don't think ceasing usage of black is necessarily a good idea. See my comments there, but I'm scared that effectively we could be pushing off new contributors with code that doesn't look standard at first glance, and if it doesn't look standard then many people might not devote the effort to contributing to us. I can say that I most likely wouldn't have started contributing to this if it weren't for me seeing the [fade to black](https://github.com/jsonpickle/jsonpickle/commit/56c44b9e541d32a35dff38bfc9ab0986f3821034) commit back in 2020 or so, as that was when I first started looking at jsonpickle.

tests/object_test.py
```diff
@@ -1140,6 +1151,21 @@ def test_datetime_with_tz_copies_refs():
...
+def test_with_exclude():
+    """ Does the _jsonpickle_exclude work"""
+    obj = ThingWithExclusion("baz")
+    encoded = jsonpickle.encode(obj)
+    decoded = jsonpickle.decode(encoded)
+    assert decoded.bar == "baz"
+    assert not hasattr(decoded, "foo")
+
+def test_contained_exclusion():
+    """ _jsonpick_exclude should work only on the class it is defined in"""
```
Theelx, line 1163: Minor typo here, should be _jsonpickle_exclude.

Theelx (changes requested):

tests/object_test.py
davvid: Agree. Please ignore my comment about single-quotes :smile_cat:

jsonpickle/pickler.py
davvid: I concur, let's defer the conversion until instantiation.

Likewise, it should be able to be specified at creation time as a parameter to `__init__`.

davvid: > then self.exclude should become self._exclude so the user doesn't mess with it after it's gone through jsonpickle's type coercion at instantiation

The only reason I vibe with for keeping it as `self.exclude` is that it lets us do without needing to have methods for adding and removing entries post-construction since our API would be `obj.exclude.add('...')`, but YAGNI so `self._exclude` is better since we can start out private by default and only expose it later if we have a strong reason to do so.

So +1 for renaming to `self._exclude` for the initial version. Plus, we can specify it at construction so there's much less need for additional API or direct attribute access. We may never actually need it.

jsonpickle/pickler.py
dwagon: Moved to use set - much better.

If you specify in the encode interface you are back to the broad brush approach, but should be easily doable.

jsonpickle/pickler.py
dwagon: Renamed to `self._excludes`

jsonpickle/pickler.py
dwagon: Agree

jsonpickle/util.py
dwagon: Agree

tests/object_test.py
dwagon: Made them single quotes anyway to be more consistent. I'm too used to black style :)

tests/object_test.py
dwagon: Fixed typo

Theelx (approved):
The changes requested were made, LGTM!

Theelx:
The test failures are from Ruff, from a previous commit having an unused variable. I'll make a PR to fix that now.

Theelx:
@dwagon If you could rebase this on the main branch which'll fix the tests, and if @davvid approves this, then I think this'll be ready to merge!

davvid (approved):

jsonpickle/pickler.py
```diff
@@ -539,6 +541,7 @@ def _flatten_obj_instance(self, obj):
         has_getnewargs_ex = util.has_method(obj, '__getnewargs_ex__')
         has_getinitargs = util.has_method(obj, '__getinitargs__')
         has_reduce, has_reduce_ex = util.has_reduce(obj)
+        self._excludes = set(getattr(obj, '_jsonpickle_exclude', ()))
```
davvid, line 544: Now that I'm re-reading this, it dawned on me that we're storing the excludes on the entire pickler instance but the excludes should probably really only affect the current object being pickled.

If y'all agree, I think we should eliminate the `self._excludes` attribute and the associated constructor argument.

What we should instead do is change this part to use a local variable `excludes = set(....)` that we can pass to `_flatten_dict_obj()`. We should change that function's signature to have an `exclude` keyword argument instead of reading it from `self._exclude`.

Another nice benefit of this is that we won't end up changing the constructor's signature.

Theelx: I think that's a good point. After all, why would we set an ``_exclude`` attribute if we're just going to overwrite it anyway (rhetorically)?

The only possible downside I can see is that for users who have to serialize lots of small objects (which is my use-case for jsonpickle in my hobby projects), having to set an extra attribute on the ``__slots__`` or ``__dict__`` for every instance could lead to a non-insignificant slowdown and especially extra memory usage, compared to setting it once on the Pickler instance. I'll fire up a benchmark real quick and edit in the results when done.

davvid:
I implemented the sugs, and rebased everything down alongside a new commit to add the changelog entry. Once @Theelx approves this seems good to merge for me.

Theelx:
Oops sorry, I see now I forgot to review this. Either way though, it looks good to me, thanks for merging davvid and thanks for the PR dwagon!

PR #518, opened 2024-07-13
ci: remove github action that creates commits

davvid:
Remove the write permission and the committing of code so that github
bots are no longer creating commits. That was an interesting setup, but
it also leads to surprises where github's branch will advance beyond
what we push.
    
Run black before isort since errors in black have a higher chance
of occuring.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,8 +1,5 @@
 name: Lint
 
-permissions:
-   contents: write
-
 on:
   push:
     branches:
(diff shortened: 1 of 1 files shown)


Theelx:
I would appreciate just using black but with their forcing of double-quoted strings enabled. I think apostrophes are more used in our text/docs anyway, so this seems annoying to have to escape all those, plus black is much more of a standard nowadays than cercis. I get disabling the CI script that auto-commits black formatting though, that can go.

davvid:
> ... especially since my muscle memory types double quotes.

It's my left pinky I try to use less to avoid RSI, but that's definitely more than fair enough given all the other reasons to stick with black.

I'll rebase to trim this down to only remove the CI commit behavior and nothing more.

After open PRs are merged I'll follow-up with a later PR to enable string normalization and format the repo. I can always type SQ and let `garden fmt` swap it over to DQ for me before committing.

For me, single quotes are clearer to read (less visual noise ~ two lines instead of one), and it meshes with how I write things when writing posix shell and perl, where SQ and DQ have different semantics, even tho that's not really a python thing. I think the other reason is that in practice I always end up using SQ because I want to embed DQ inside the the string in error and info/debug messages without using backslash escapes, but that's just a subset of strings. I've personally considered PSF's attitude around black problematic, but that's just me and it's just a personal opinion.

I once opened a PR against black trying to add an option to normalize to SQ (which went down exactly as one would expect), and I think that commit ended up in cercis :rofl: so I have a personal soft spot for it, but that's not reason enough.

Thanks for reviewing. I just pushed up a commit that trims this PR down. I'll update the title/description shortly.

Theelx:
You definitely raise a good point about meshing with POSIX shells and perl, especially since we do use bash scripts in this project. Though this PR might not be the best place to discuss it, I think it would be worth getting opinions from users as to whether they prefer DQ or SQ. I'm going to create an RFC issue shortly to gather the arguments for/against DQ and SQ, so that we can get a fuller picture of the userbase's preferences.

Regardless, thank you for making this PR, and raising the issue with GitHub advancing past where we push! I think it looks good to merge, as the test errors seem to be with a newer version of ruff being used, and that should be fixed in a follow-up PR.

PR #523, opened 2024-09-19
Include how to run examples in README

Theelx:
This PR just includes some extra info about how to run examples in the examples/README.md file. I didn't include full instructions for Mac because it should be similar to Linux, nor non-Ubuntu Linux distros because I don't use them. Also the Windows instructions simply tell the user to follow online guides for installing Python and Git as many exist, but at some point I want to include short directions inside this file, This PR should be squashed when merged.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,4 +1,4 @@
-collect_ignore = ['contrib']
+collect_ignore = ['contrib', 'examples']
 
 
 def pytest_addoption(parser):
--- a/examples/README.md
+++ b/examples/README.md
@@ -2,3 +2,39 @@
 This directory exists to provide examples of how to use jsonpickle to achieve simple goals, such as:
 - Saving instances of a class to a file
 - And more to come!
+
+## Running the code
+### Linux (Ubuntu)
+If you don't have python and/or git installed, install those first:
+```sh
+apt install git python3 python3-dev python3-pip
+```
+Once you have those installed, start by cloning the repository and entering into the directory (or copying the file locally):
+```sh
+git clone https://github.com/jsonpickle/jsonpickle.git && cd jsonpickle/examples
+```
+Then, you should install jsonpickle:
+```sh
+pip install jsonpickle
+```
+Lastly, you can run any example file as long as you have jsonpickle and Python installed!
+```sh
+python3 save_class_to_file.py
+```
+You can also open the .py file in your favorite text editor, such as VSCode, gedit, or others!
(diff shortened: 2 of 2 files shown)


davvid:
Very nicely done. My only sug (and this is a tough one because it exposes extra complexity, although it does help users down the line) might be to suggest the use of a virtualenv.

I've had to help users unbreak their setup in the past because `pip install` drops files in their global $HOME directory so encouraging its use here might lead them to install other libraries (with more dependencies) using the same method.

I would suggest this recipe instead:

```bash
python -m venv env
./env/bin/pip install jsonpickle
./env/bin/python save_class_to_file.py
```

Alternatively, we could suggest use of `./env/bin/pip install --editable ..` so that it installs directly from the git clone.

If we want to avoid the `./env/bin/...` stuff then we could instead suggest the "activate" workflow:

```bash
# Create a virtualenv in the "env" directory. You don't need to run this again.
python -m venv env

# You must activate the virtualenv anytime you open a new shell in order to
# make the to-be-installed jsonpickle package available.
source env/bin/activate

pip install jsonpickle
python save_class_to_file.py
```

The virtualenv workflow is more complicated because it brings along additional filesystem and per-shell environmental state, but it's generally a good python practice since it avoids global $HOME or system state so in my opinion the pros outweigh the long-term cons.

Let me know what you think. I wouldn't mind if this were merged as-is if you disagree, though, since in practice these `pip` invocations will only install jsonpickle, and jsonpickle has no required dependencies outside of the stdlib.

davvid (approved):

Theelx:
> I would suggest this recipe instead:

I think it may be worse to use that recipe, but only because it requires knowledge of PATHs for a user to understand what's happening, because it will seem to differ between Linux/Windows for users used to Windows backslashes (though I believe Windows still accepts forward-slashes), and most importantly because it's not common/it doesn't build the kind of muscle memory that most beginners seem to need nowadays. I don't think I've ever seen the use of ``./env/bin/pip`` before, so I don't know whether it has the same effect as source though. I do agree that recommending a virtualenv by default though would be a good idea, so I think your suggestion including ``source env/bin/activate`` is likely more intuitive for the people who'll need this guide.

I'll add your second suggestion in as the preferred option for Linux, but include the current version as an alternative. It seems ``source`` works differently on [Windows](https://stackoverflow.com/q/8921188), so I'll put `` activate`` as the option for Windows.

PR #524, opened 2024-09-23
Support 3 12 & Update GHA Versions

Theelx:
This PR is relatively straightforward, just adding support for Python 3.12 in the testing matrix as well as updating the versions of various actions used in the test/lint workflows. While they could be two separate PRs, I did them in one because they both touch GitHub Actions, and because I think of them as both updating versions. The only thing I'm unsure about is whether gmpy2 should be explicitly required to be at least 2.2.0 for Python 3.12 or not, as technically 2.2.0 is the first version that supports 3.12/3.13.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,11 +10,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
-      - uses: actions/setup-python@v4
+      - uses: actions/setup-python@v5
         with:
-          python-version: "3.10"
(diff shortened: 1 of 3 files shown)


davvid:
Adding a constraint for python3.12 seems like it might be a good idea, particularly if the unconstrained gmpy2 runs into errors due to pip trying to use some older incompatible version. We can always tweak constraints later if we start to run into gmpy2 versioning issues.

That said, I think pip will try to use the latest compatible version so it might not make a difference in practice since pip will already try to use a fairly new version.  We probably would benefit more from running our tests against a continually-latest gmpy2 version instead of constraining it in the long run.

So, I think this looks good as-is (unconstrained) if it's passing w/out errors already.

PR #526, opened 2024-10-09
Improve benchmarking commands

Theelx:
With my newly acquired desktop computer and its increased core count (4->16 w/o SMT) I now have to worry more than before about core affinity. By default, Linux will switch tasks between cores every so often, which can massively increase outlier times on some tests. Pinning the task to one core, as well as disabling python's GC mechanism, reduces the maximum outlier times from 200x normal on some tests to a maximum of around 10-15x normal, which leads to markedly closer mean/median times. Additionally, with the recent release of Python 3.13, users now have the option to enable CPython's experimental JIT, which need warmup iterations to warm up the JIT when enabled. While the JIT is off by default at the moment, future releases sound like they may enable it by default, which would effectively require benchmark warmups, so I'm enabling it here to stay ahead of the curve ;)


Theelx:
Later on, I plan to add automatic checking of whether taskset is installed on the system, but that's annoying to do in a makefile so I haven't implemented it _quite_ yet.

davvid (changes requested):
> [...] but that's annoying to do in a makefile so I haven't implemented it quite yet.

In case it helps, it might be slightly easier to implement it in the  `garden.yaml` file since. The syntax is a bit easier compared to makefiles since you only need to know shell scripts + yaml.

Thanks for these updates, good stuff.

We do have a `garden bench` command implemented in the current garden file so we'll probably want to update it as well for parity with the new changes here. I can take a stab at keeping it in sync with these changes later after this is merged if you don't get a chance to update it.

Theelx:
> > [...] but that's annoying to do in a makefile so I haven't implemented it quite yet.
> 
> In case it helps, it might be slightly easier to implement it in the `garden.yaml` file since. The syntax is a bit easier compared to makefiles since you only need to know shell scripts + yaml.

I will do that!
> Thanks for these updates, good stuff.
> 
> We do have a `garden bench` command implemented in the current garden file so we'll probably want to update it as well for parity with the new changes here. I can take a stab at keeping it in sync with these changes later after this is merged if you don't get a chance to update it.

Got it, I'll update that. I found a good guide for reproducible benchmarking on Linux a few days ago that I have been trying to implement on my desktop (it's slow because it requires changes to kernel parameters sadly). I'm not planning to merge this until I can integrate some of the lessons from that guide into this anyway.

davvid (approved):

Theelx:
I ended up having to implement it in the makefile, as its fairly complex logic would completely destroy garden's readability. I'm currently making a README file for the benchmarking/ directory to explain how to use everything.

Theelx:
The ``make benchmark`` command uses different syntax than the rest of the commands because I couldn't figure out how to do conditionals and stuff cleanly without getting syntax errors. If you have no objections to this, I'll go ahead and merge, but otherwise I'd like to hear any issues since I've made fairly big changes since your last review.

davvid (approved):
I would suggest moving the sudo install steps to a note in the benchmarking/README, but we can always do that later if you agree. This is good for merging as-is.

For shell scripting, I usually follow git's guide which avoids some idioms like `[` and `]` in favor of a more regular syntax using `test`.

For example,

```bash
if test -f /etc/os-release; then ...
if test "$$ID" = ubuntu; then ...
if ! type taskset >/dev/null 2>&1; then ...
```

You mention garden's syntax getting complicated. I might play around with that later to see what you mean, so I'm curious to hear more details if you've got 'em.

I'll start merging merging this one and the current batch of the ones I submitted now since these are all good to go. Thanks for reviewing these!

PR #527, opened 2024-10-13
packaging: replace deprecated setup.cfg with pyproject.toml

davvid:
Update to the modern practice of a using a single declarative
pyproject.toml file for package configuration.
    
`setup.cfg` is deprecated and is no longer necessary.

Ref: https://github.com/pypa/setuptools/issues/3214

Includes related updates to the optional dependencies and
garden commands in separate commits.


davvid:
This will conflict with the `CHANGES.rst` edits in #514 so we'll merge that first and then I'll rebase and resolve.

Theelx (approved):
I see no problems with this in its current state!

PR #528, opened 2024-10-13
backend: add "yaml" as a supported backend

davvid:
YAML is a superset of JSON and has many nice features such as
support for comments and a generally nicer syntax.

Add support for using "yaml" as a jsonpickle backend.


davvid:
This uses changes from #527 so I'll rebase this after that's merged to make the diff in this PR easier to review.

jsonpickle/backend.py
```diff
@@ -16,11 +16,7 @@ def _verify(self):
         """Ensures that we've loaded at least one JSON backend."""
         if self._verified:
             return
-        raise AssertionError(
-            'jsonpickle requires at least one of the '
-            'following:\n'
-            '    python2.6, simplejson'
-        )
+        raise AssertionError('jsonpickle could not load any json modules')
```
Theelx, line 19: Thanks for updating this!

jsonpickle/backend.py
```diff
@@ -98,6 +94,9 @@ def __init__(self, fallthrough=True):
 
         self.load_backend('simplejson')
         self.load_backend('json')
+        self.load_backend(
```
Theelx, line 97: If I recall correctly, aren't backends attempted in the order in which they are loaded? Might make sense to load ujson first, at least until people start using the yaml backend.

Theelx (approved):
Minor nitpick about yaml/ujson backend priority, but otherwise LGTM after the rebase!

jsonpickle/backend.py
davvid: True, sorry for not catching this until post-merge. I'll fix this up locally and push it out shortly.

Issue #407, opened 2022-09-30
pandas dataframe with list as values


feizerl:
Consider this code:

```py
import jsonpickle
import jsonpickle.ext.numpy as jsonpickle_numpy
import jsonpickle.ext.pandas as jsonpickle_pandas
import pandas as pd

jsonpickle_pandas.register_handlers()
jsonpickle_numpy.register_handlers()


a = pd.DataFrame({"date": [["20220921"]]})
b = jsonpickle.decode(jsonpickle.encode(df, keys=True))
```

if you run this, you get:

```py
$ py -i foo.py 
>>> a['date'][0]
['20220921']
>>> b['date'][0]
"['20220921']"
```

i.e. a `list[str]` has been changed to a `str`

Is there any way to avoid this?

Theelx:
Update: This is definitely a bug, and it'll be hard to fix. Pandas doesn't have a string dtype, only an object dtype, so it's hard for jsonpickle's pandas extension to differentiate between non-integer types, such as a list and a string. I'm working on fixing the encoding for that, but it'll be difficult as I need to integrate type() checking too, in addition to dtype checking.

Theelx:
Update 2: I just realized this'll be harder to fix than I previously thought, since the lists can contain more than one dtype. For example, one could have a dataframe like so: ``pd.DataFrame({"date": [["20220921", 20220921]]})``, which has string and int dtypes inside the list. This means we'll have to store the dtypes for every single item inside the list, which is going to massively bloat the encoded blob. I'm reconsidering whether it's worth fixing this bug in the general case of a multi-dtype list.

Theelx:
Apparently this is basically the same issue as #358.

Paradoxdruid:
I wanted to use jsonpickle for a bioinformatics project of mine, but this behavior is a killer; I regularly have dataframes with lists of floats.

Could an implementation like this, which stores lists as lists of tuples of (value, type), be useful?  It roughly doubles the size of the representation of lists as values in pandas Series, but wouldn't touch elsewhere, I think.

```python
# a naive implementation
def list_flatten(list_input: List[Any]) -> str:
    nested_list = [f"({each}, {type(each).__name__})" for each in list_input]
    return json.dumps(nested_list)

def unflatten_list(str_input: str) -> List[Any]:
    obj = json.loads(str_input)
    final_list: List[Any] = []
    for each in obj:
        value, str_type = each.split(",")
        str_type = str_type.strip()
[...]

Issue #358, opened 2021-06-09
Pandas list value not decoding properly


brandonschabell:
When decoding a Pandas Dataframe which contains a list value, it is being decoded as a string.

```python
import jsonpickle
import jsonpickle.ext.pandas as jsonpickle_pandas

jsonpickle_pandas.register_handlers()

df = pd.DataFrame({'col1': [[1, 2, 3], ['a', 'b', 'c']]})
assert type(df['col1'].iloc[0]) == list
data = jsonpickle.encode(df)
result = jsonpickle.decode(data)
assert type(result['col1'].iloc[0]) == list  # AssertionError: assert <class 'str'> == <class 'list'>
```

losito105:
Hello! I'm just getting started with open source but I've used jsonpickle before and I'd like to attempt to resolve this issue.

losito105:
I am new to open source contribution, but I've worked with both jsonpickle and pandas in the past and I'd like to try to tackle this issue.

PR #534, opened 2024-11-03
Add support for nested python containers and mixed dtypes

### LINKED ISSUE #457: Pandas series with object type get converted to strings
This error is a slightly different take on https://github.com/jsonpickle/jsonpickle/issues/407 and https://github.com/jsonpickle/jsonpickle/issues/358.

My issue is that if a dataframe has a column of mixed type (e.g. `[4, 'foo']`), then it will be converted to strings on unpickling (e.g. `['4', 'foo']`):

```py
import pandas as pd
import jsonpickle as jp
import jsonpickle.ext.numpy as jsonpickle_numpy
import jsonpickle.ext.pandas as jsonpickle_pandas
jsonpickle_numpy.register_handlers()
jsonpickle_pandas.register_handlers()

# Create simple data frame of mixed data types
df1 = pd.DataFrame((
  a = [1, 2], 
  b = [4, 'foo']
))

# Convert to jsonpickle
string = jp.dumps(df1)
df2 = jp.loads(string)

# Show that it didn't work
print(df1.b.values)
print(df2.b.values)

assert df1.b[0] == df2.b[0] # False: the first is 4 and the second is '4'
```

I know there isn't an easy fix for this, but I'm hopeful that there's _some_ fix, since (at least in this case) the actual dataframe data looks a lot like JSON format already! In particular, one suggestion I have is that currently just a single `dtype` is stored per column:
```py
"meta": {"dtypes": {"a": "int64", "b": "object"}
```

My suggestion that I believe would solve this particular bug would be that if the `dtype` is `object`, to store a list (one for each element) instead. For example, in this case:
```py
"meta": {"dtypes": {"a": "int64", "b": {"object": ["int", "object"]}}
```

Personally, I also think storing `values` as a dict of lists (rather than as a string) would be more robust _and_ easier to read/interpret. I feel `pd.DataFrame.to_dict()` is already pretty close to what would be required!

Theelx:
Fixes #457, #407, and #358 (and potentially more that I don't want to bother to find). Effectively, this PR adds a bunch of code in order to properly preserve pandas dtypes over encode/decode roundtrips. I've now spent roughly 20 hours working on this over 3 different attempts, so I'll need to take a short break before addressing any review comments (though it's quite possible that due to the amount of churn any review will take a long time)!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,11 @@ Upcoming
     * ``yaml`` is now supported as a jsonpickle backend. (+528)
     * `OSSFuzz <https://github.com/google/oss-fuzz>`_ scripts are now available in
       the ``fuzzing/`` directory. (+525)
+    * Pure-python dtypes are now preserved across ``encode()``/``decode()`` roundtrips
+      for the pandas extension. (#407) (+534)
+    * Pandas dataframe columns with an ``object`` dtype that contain multiple different
+      types within (e.g. a column of type ``list[Union[str, int]]``) now preserve the types
+      upon being roundtripped. (#457) (#358) (+534)
 
 v3.3.0
 ======
--- a/jsonpickle/ext/pandas.py
+++ b/jsonpickle/ext/pandas.py
@@ -1,15 +1,62 @@
+import warnings
 import zlib
 from io import StringIO
 
 import pandas as pd
+import numpy as np
 
 from .. import decode, encode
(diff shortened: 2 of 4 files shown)


davvid (approved):

Theelx:
Oops, it just occurred to me that we'll need a helper to decode objects created with previous version of the extension! I'll make that real quick.

PR #537, opened 2024-11-06
API: remove breaking changes for v3.4.2

davvid:
v3.4.1 should have been released from the "maint/v3" branch but it was released from "main" instead, which contained changes that were meant to be deferred until v4.

Revert these changes to prevent these from getting released for now, and so that we can patch v3.4.2 without the breaking changes.


davvid:
Sorry for leaving this footgun in the tree and not making it clear that any v3 releases would need to be done from the maint/v3 branch.

Once you've seen this we can deploy v3.4.2 and we should probably yank v3.4.1 from pypi, as these changes were intended for v4.

PR #538, opened 2024-11-06
v4.0.0

davvid:
jsonpickle v4.0.0

PR #539, opened 2024-11-08
tests: datetime.datetime.utcnow() is deprecated

davvid:
Use datetime.datetime.now(tz=datetime.timezone.utc) instead.


davvid:
I think we should also merge this before v4 #538.

Theelx (approved):
Looks good!

PR #541, opened 2024-11-17
api: deprecate the jsonpickle.compat module

davvid:
This module is no longer needed now that we're Python 3-only.

Keep the module around for compatibility but remove all internal references.

### DIFF
--- a/jsonpickle/util.py
+++ b/jsonpickle/util.py
@@ -19,12 +19,12 @@ import types
 from collections.abc import Iterator as abc_iterator
 
 from . import tags
-from .compat import numeric_types
 
+NUMERIC_TYPES = (int, float)
 ITERATOR_TYPE = type(iter(''))
 SEQUENCES = (list, set, tuple)
 SEQUENCES_SET = {list, set, tuple}
-PRIMITIVES = {str, bool, type(None)} | set(numeric_types)
+PRIMITIVES = {str, bool, int, float, type(None)}
 FUNCTION_TYPES = {
     types.FunctionType,
     types.MethodType,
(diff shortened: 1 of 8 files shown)


jsonpickle/util.py
```diff
@@ -16,13 +16,15 @@
...
 import time
 import types
+from collections.abc import Iterator as abc_iterator
 
-from . import compat, tags
-from .compat import abc_iterator, class_types, iterator_types, numeric_types
+from . import tags
 
+NUMERIC_TYPES = (int, float)
+ITERATOR_TYPE = type(iter(''))
```
davvid, line 24: These two symbols have been moved into the `utils` module.

Alternatively, I could make these private so that we don't increase the API surface.

Another thought I had is that we could inline the `(int, float)` tuple into the small handful of call sites where it's checked, but I kept it here since we use the symbol in a few different files.

If we think shrinking the API surface is worth it then I'd be okay inlining it into the calling modules. Let me know what you think about that these.

davvid: Also, reading through this code made me realize that the code paths that use `ITERATOR_TYPE` might need some attention. It seems like that expression returns a very specific type of iterator, specifically `<class 'str_ascii_iterator'>`, and I suspect that our code was hoping for a common base class of all iterators.

Unrelated to these changes, but just something that came to mind so I figured I'd write it down before I forget. Hope you've been well, cheers.

jsonpickle/util.py
```diff
@@ -16,13 +16,15 @@
...
 SEQUENCES = (list, set, tuple)
 SEQUENCES_SET = {list, set, tuple}
-PRIMITIVES = {compat.ustr, bool, type(None)} | set(numeric_types)
+PRIMITIVES = {str, bool, int, float, type(None)}
```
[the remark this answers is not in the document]
davvid: Good idea. I'll inline all of these and avoid adding additional module globals. Less is more.

PR #545, opened 2024-12-13
jsonpickle: guard against invalid b85 and b64 data

davvid:


davvid:
I'll add a changelog entry tonight once I'm back home.

davvid:
I found a few more out-of-band values that could trigger other errors so I amended this commit with  more tests and exception cases to cover these values for b64 and b85. We have a number of unreleased fuzz fixes at this point so I'm going to deploy a patch release today.

PR #546, opened 2024-12-20
unpickler: increase robustness to invalid py/id data

davvid:
Detect TypeError when py/id values are resolved so that we can return None when the payload contains invalid data.

### DIFF
--- a/jsonpickle/unpickler.py
+++ b/jsonpickle/unpickler.py
@@ -313,8 +313,15 @@ def _loadmodule(module_str):
 
     """
     module, identifier = module_str.split('/')
-    result = __import__(module)
-    for name in identifier.split('.')[1:]:
+    try:
+        result = __import__(module)
+    except ImportError:
+        return None
+    identifier_parts = identifier.split('.')
+    first_identifier = identifier_parts[0]
+    if first_identifier != module and not module.startswith(f'{first_identifier}.'):
+        return None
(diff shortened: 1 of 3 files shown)


davvid:
I just added a 2nd commit for handling invalid py/repr data as well.

jsonpickle/unpickler.py
```diff
@@ -917,7 +929,9 @@ def _restore_dict(self, obj):
                 else:
                     str_k = k
                 self._namestack.append(str_k)
-                data[k] = self._restore(v)
+                data[k] = result = self._restore(v)
+                if isinstance(result, _Proxy):
```
Theelx, line 933: Is this fixing something found in the fuzzing or was the lack of it just causing a bug of some other sort?

Theelx (approved):

jsonpickle/unpickler.py
davvid: Yup, this edge case was found while fixing the  fuzzing errors. We end up leaving `_IDProxy` objects in the final output without this extra check.

This probably isn't too great from a performance perspective but we'd probably need some deeper restructuring to handle all of the edge cases in a more performant manner. e.g. either never return _IDProxy, or change the `def _restore(...)` return value to return `(object, bool is_proxy)` tuples to indicate that a proxy is being returned.

Issue #550, opened 2025-02-17
Surprising behavior loading invalid JSON


jaraco:
I have [some tests](https://github.com/jaraco/jaraco.abode/blob/d855d9cfbdf529e9a9314964f0a62ba7a1f38783/tests/test_abode.py#L514-L517) that create invalid json (`invalid json`) and attempt to parse it to validate lenient behavior over a corrupted jsonpickle file.

I was surprised to learn that `invalid json` actually decodes as a JSON pickle if a YAML library is present:

```
 🐚 pip-run jsonpickle pyyaml -- -c 'import jsonpickle; print(jsonpickle.decode("invalid json"))'
invalid json
```

That file is clearly invalid JSON, but because jsonpickle falls back to other encoders, including YAML, by default, it attempts to parse that string as YAML, which is lenient and treats many strings as literal strings.

The fallback to other JSON libraries makes sense, but I'm thinking the fallback to YAML (and maybe other json libraries) should be opt-in, not opt-out.

davvid:
The yaml backend was only recently added so I wouldn't mind making it opt-in. We can treat it similarly to the numpy support that's provided through a `jsonpickle.ext` module.

My sug for what the opt-in would look like would be:

```python
from jsonpickle.ext import yaml

yaml.register()
```

The signature would be `def register(backend=None)`. When `backend` is `None` it'd default to registering the yaml module with global `jsonpickle.backend.json` `JSONBackend` instance. This lets us supply our own JSONBackend instance if we don't want to register yaml with the global backend instance.

If that sounds good to you and @Theelx I can take a stab at making it work this way. 

As far as deploying this change goes, it seems like this should be a minor semver bump at minimum. I'd suggest that it should be a major version bump, but I'll defer to you two for your opinion on whether we shouldn't bump the major version. Although the vast majority of users will be unaffected by this change, we don't have any way of knowing whether anyone is relying on the yaml backend's presence so it's better to be safe IMO. I'm leaning towards a major version bump for this proposed change.

In the meantime, we currently have three unreleased fixes in our git tree so I'll plan to tag and deploy `v4.0.2` shortly to clear the road for these proposed changes.

davvid:
`v4.0.2` has been deployed. #551 is the proposed update to make yaml opt-in.

jaraco:
[...]

PR #551, opened 2025-02-17
backend: make the yaml backend opt-in

davvid:
YAML is a much more lenient format so it can cause jsonpickle to load files that are expected to fail when interpreted using a JSON reader. Let users control whether or not the yaml backend is used.

Move registration of the yaml backend into a separate jsonpickle.ext.yaml module. Do not register the yaml module by default so that users can opt-in to it.

Closes: #550
Reported-by: Jason R. Coombs @jaraco

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+Upcoming
+========
+    * **Breaking Change**: The ``yaml`` module is no longer registered by default.
+      You can re-enable yaml support using ``jsonpickle.ext.yaml.register()``.
+      (#550) (+551)
+
 v4.0.2
 ======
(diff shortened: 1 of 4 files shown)


jaraco (approved):

Theelx:
I'm fine with this, but I'd like to see if we can combine a few other breaking changes with this before releasing 5.0.0 especially because we're only on 4.0.2. I'll go through the backlog to see if there's anything else that can be easily included. Also, some documentation for this being an extension outside of the changelog would be nice, I'll try to make some within the next few days.

Theelx (approved):
Code looks good for now, I'll make some more documentation in a follow-up PR. I agree that this needs to be a major bump but let's see if we can fit anything else in v5 before releasing it.

Issue #555, opened 2025-03-18
Round-trip numpy.datetime64 at nanosecond resolution


bitbckt:
jsonpickle is unable to round-trip a nanosecond resolution numpy.datetime64. Microsecond resolution used as a working example in a REPL session:

```
>>> import numpy as np
>>> import jsonpickle
>>> import jsonpickle.ext.numpy as jsonpickle_numpy
>>> jsonpickle_numpy.register_handlers()
>>> jsonpickle.decode(jsonpickle.encode(np.datetime64("2025", "us")))
numpy.datetime64('2025-01-01T00:00:00.000000')
>>> jsonpickle.decode(jsonpickle.encode(np.datetime64("2025", "ns")))
```

Throws `ValueError: Converting an integer to a NumPy datetime requires a specified unit`.

Theelx:
Thanks for reporting this! I'll look into it shortly, I think it's related to a recent change from 4.0.

davvid:
I have a possible solution for this. I'll post it shortly for your review @Theelx

bitbckt:
Thanks for fixing this! Is there a release coming up that I can look forward to?

Theelx:
We hope to get 5.0.0 out soon (hopefully within a few weeks). I've been super busy with schoolwork lately though, so I can't guarantee that we'll be able to get it out soon because we were hoping to fit a bunch of the backwards-incompatible changes that need to be made into it.

davvid:
Do you think it'd make sense for me to create a `maint/v4` branch for the purposes of creating a v4 patch release without the yaml opt-in breaking change?

If not, my other thought was that maybe I can create a "v5.0.0.pre.0" pre-release tag (may not be the same exact naming -- I have to read up on `setuptools-scm` and its tag naming conventions) so that folks can try out these changes in a preview / pre-release form before we do the official release.

I think maybe a pre-release tag might be the most helpful so I'll be reading up on that later this week. Shout out if you think that's not a good idea and I'll refrain from doing so.

I'll ping back here if I do end up creating a pre-release. No promises, though ~ it's just a thought / consideration at this point that *seems* like a useful thing to start doing at times.

Theelx:
Yes, I think a maint/v4 branch would be reasonable. You can also do a pre-release, I don't think that could hurt!

davvid:
`v4.0.3` has been released with a backport of this fix.

https://pypi.org/project/jsonpickle/4.0.3/

PR #556, opened 2025-03-19
numpy: add support for nanosecond-resolution datetime64 objects

davvid:
Extend NumpyGenericHandler to special-case datetime64 dtypes with nanosecond resolution. This handling seems to only be needed for nanosecond resolution.  Other units do not require special handling.

Closes: #555


Theelx (approved):
Looks great!

PR #557, opened 2025-03-19
github: run actions on ubuntu-latest

davvid:


davvid:
The CI tests were failing when I submitted #556.

I recall taht github recently updated their runners so I changed our config to use ubuntu-latest and the errors went away.

I'll merge this change and then rebase the other PR shortly.

Issue #558, opened 2025-03-29
setuptools deprecation warnings


mgorny:
As of 4.0.3, building the wheel produces the following warnings:

```
/usr/lib/python3.10/site-packages/setuptools/config/_apply_pyprojecttoml.py:55: SetuptoolsDeprecationWarning: 'tool.setuptools.license-
files' is deprecated in favor of 'project.license-files' (available on setuptools>=77.0.0).
!!

        ********************************************************************************

        By 2026-Feb-18, you need to update your project and remove deprecated calls
        or your builds will no longer be supported.

        See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-files for details.
        ********************************************************************************

!!
  _apply_tool_table(dist, config, filename)
/usr/lib/python3.10/site-packages/setuptools/config/_apply_pyprojecttoml.py:61: SetuptoolsDeprecationWarning: License classifiers are d
eprecated.
!!

        ********************************************************************************
        Please consider removing the following classifiers in favor of a SPDX license expression:

        License :: OSI Approved :: BSD License

        See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
        ********************************************************************************

!!
  dist._finalize_license_expression()
/usr/lib/python3.10/site-packages/setuptools/dist.py:759: SetuptoolsDeprecationWarning: License classifiers are deprecated.
!!

        ********************************************************************************
        Please consider removing the following classifiers in favor of a SPDX license expression:

        License :: OSI Approved :: BSD License

        See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
        ********************************************************************************

!!
  self._finalize_license_expression()
```

davvid:
Thanks for the heads-up. That's annoying. Something so core like setuptools should probably reconsider deprecating stuff that's working just fine for no real benefit.

The last time I updated my Debian/testing environment it updated me to setuptools 44.1.1 and it tells me that I can upgrade to setuptools 75.0.0 based on what they have available in their repos today.

I tested things and it's not as simple as moving the `license-files` field from `tools.setuptools.license-files` over to `project.license-files`. That actually ends up breaking environments with older setuptools versions so it's a non-starter.

Luckily, it does look like we do have a straightforward solve for this situation. We can simply omit this field altogether. This avoids the deprecation while also doing the right thing because setuptools finds the license file automatically because it knows to look for a file called `LICENSE`, which we already provide.

[...]

PR #559, opened 2025-03-29
Add SPDX license identifier

cdce8p:
Followup to #558

Unfortunately, removing the license classifier in https://github.com/jsonpickle/jsonpickle/commit/ae9505871a92a0b7595687b677d1b9b25df45e58 means that PyPI won't show any license information anymore. Ideally it should not be removed until the PEP 639 syntax with license expression is adopted. This was only added in setuptools `77.0.3` which requires Python 3.9, so sadly incompatible at the moment.

I'd suggest to use the SPDX license identifier with the legacy `project.license.text` field instead, until Python 3.8 is dropped. _Note that this will emit a DeprecationWarning until then, probably something which should be relaxed in setuptools._

https://spdx.org/licenses/BSD-3-Clause.html

/CC @davvid

| before (4.0.3) | after (4.0.4) |
| ------ | ----- |
| Screenshot 2025-03-29 at 16 09 42 | Screenshot 2025-03-29 at 16 10 04 |

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,6 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "jsonpickle"
+license = { text = "BSD-3-Clause" }
 authors = [
     { name = "Theelx" },
     { name = "David Aguilar", email = "davvid+jsonpickle@gmail.com" },


davvid:
> Note that this will emit a DeprecationWarning until then, probably something which should be relaxed in setuptools.

100% agree on the sentiment that this DeprecationWarning should be relaxed in setuptools. I personally see no point in making this a breaking change because it bifurcates the ecosystem and adds friction for library authors that maintain support for a wide range of environments.

I tested this and, yup, this does keep things happy for older pre-77 setuptools. It does bring back the deprecation warning for newer setuptools (which was the motivation for the original change) which is mildly annoying.

I have one tiny sug, though: please rebase this change and have it target the `maint/v4` branch so that we can target this change against a v4 patch release.

---------------------------------------------------------

I should probably find the right venue to give constructive feedback to the setuptools authors so that they can tell me that I'm wrong and that we shouldn't try to keep compatibility with older setuptools versions :smile:. @jaraco if you're listening I'm all ears if you have guidance on how to proceed in the future. Luckily it's "just" a deprecation warning at this point.

Separate from this change, I suspect that we should bump up the minimum setuptools version that we have specified in `pyproject.toml` to `>= 65.5` (or some other version?) on the `main` branch. We haven't tagged the v5 final release yet so that's something we can include alongside the breaking change that we have sitting in the queue.

cdce8p:
> I have one tiny sug, though: please rebase this change and have it target the `maint/v4` branch so that we can target this change against a v4 patch release.

Done. Wasn't sure which branch I should target initially which is why I opened it against `main`.

> I tested this and, yup, this does keep things happy for older pre-77 setuptools. It does bring back the deprecation warning for newer setuptools (which was the motivation for the original change) which is mildly annoying.

Just technically speaking, it would be possible to add an upper bound for setuptools, say `<77`. With that the Deprecation Warning wouldn't show up.

It might be a good time now to admit that I was one of the people involved with the PEP 639 implementation in setuptools, so the warning is partially my own fault. 😅 Anyway after thinking some more about it, in the end the warning is correct. It's just unfortunate that it even shows up for project which can't / don't want to use `>=77` yet. Not quite sure there is much that can be done about it though.

As for breaking the ecosystem: I don't suspect it will be changed to an error anything soon. _Though that's out of my control, I guess._ It's more a warning to nudge the few users who actually see it towards using the new syntax.

PR #560, opened 2025-04-16
Fix test_multindex_dataframe_roundtrip on 32-bit architectures

cjwatson:
https://bugs.debian.org/1103146 reports that this test is failing on Debian's i386 architecture, and
https://ci.debian.net/packages/j/jsonpickle/testing/armel/, https://ci.debian.net/packages/j/jsonpickle/testing/armhf/, and https://ci.debian.net/packages/j/jsonpickle/testing/i386/ show that it has been failing on all 32-bit architectures for some time.  The reason is buried many levels deep, but it turns out that
`pandas.core.indexes.base.Index._with_infer` converts the index levels to `np.int64` even if they were initially of a narrower integer type - so it seems to make most sense to start them out as `np.int64` as well.

### DIFF
--- a/tests/pandas_test.py
+++ b/tests/pandas_test.py
@@ -99,7 +99,7 @@ def test_multindex_dataframe_roundtrip():
     df = pd.DataFrame(
         {
             'idx_lvl0': ['a', 'b', 'c'],
-            'idx_lvl1': np.int_([1, 1, 2]),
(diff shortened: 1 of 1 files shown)


Theelx:
Yep this makes sense. Thanks for finding the issue and sending the PR, I'll merge it now!

PR #562, opened 2025-05-27
Fix legacy pandas dataframe decoding

Theelx:
This fixes an issue I discovered when typing the pandas extension where we'd get a JSONDecodeError on PandasDfHandler's restore. I also added a warning and a test, and this test adds coverage for the make_read_csv_params function (which was previously untested). The make_read_csv_params function is only used for the pre-3.4 decoding though, so it's just a nice-to-have in terms of coverage rather than essential, because the restore_v3_3 function (which is the only caller for make_read_csv_params) should never change anymore. This should be released in its own 4.1.1 release if approved by @davvid (I would appreciate if you could do the release davvid).

One thing I'm considering is whether the two neighboring try/excepts should be merged, the only issue I can think of though would be the loss of specificity for the error messages (as to whether we know they're decoding a pre-3.4 object or not). Not sure how big of a deal this would be, but I'm leaving it separate for now.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@ Upcoming
       You can re-enable yaml support using ``jsonpickle.ext.yaml.register()``.
       (#550) (+551)
 
+v4.1.1
+======
+    * An error in the jsonpickle pandas extension when decoding objects that were encoded
+      before jsonpickle v3.4.0 was fixed, and warnings were added. (+562)
+
 v4.1.0
 ======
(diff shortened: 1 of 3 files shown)


davvid (approved):
Thanks, this looks good to me. I'll keep an eye on this and once you merge I'll tag and publish `v4.1.1`.

Theelx:
@davvid Can you tag and publish 4.1.1?

davvid:
I updated the `maint/v4` branch to omit the upcoming breaking change from v5.

https://pypi.org/project/jsonpickle/4.1.1/

Issue #554, opened 2025-03-17
Add Type Stubs


donBarbos:
Hi!
I noticed that you don't use type hints in your code, so I have a suggestion to add type stubs.
What do you think about it?

Theelx:
Hello, we would love to implement them but we haven't had the time yet. If you'd be able to send a PR implementing them then that'd be greatly appreciated. I do think we'd rather have inline type hints rather than a py.typed or .pyi file though.

PR #563, opened 2025-06-04
V5: Add type hints, drop support, remove functions

### LINKED ISSUE #441: Would it be possible to add a stubfile?
Hey all, 

I've started using typing in python and have managed to get literally all my packages working with pylint, with the exception of jsonpickle. 

Screenshot 2023-03-31 at 4 08 57 PM
Screenshot 2023-03-31 at 4 09 32 PM

Any chance we could get a stub file included?

### LINKED ISSUE #561: RFC: Misc Typing Work and Deprecations for 5.0.0
Hello everyone, I'm making this public instead of private correspondence between me and davvid so that any potential contributors can see our thought process. Anyway, to get straight into it, I'm working on adding inline type hints to the code with MonkeyType and pytest-monkeytype in preparation for 5.0.0. I just started an hour or two ago and I've now manually gone over all of its suggestions for jsonpickle.util and I found a few interesting things:
1. There are a lot of one/two-line functions in jsonpickle.util that are used in only one or two places or only in tests/jsonpickleJS (e.g. is_not_class, is_sequence, is_bytes, is_tuple, and more). I think that for 5.0.0 it would be reasonable to just drop these functions and inline them since Python isn't smart enough to optimize the function calls out. Not that performance is super important for some of these, however it'd still be nice for removing some unused/barely-used code. If we do that, should we include a deprecation period by emitting a deprecation warning for a month or two of the 4.x.x series, or is that unnecessary for such simple functions?
2. I'm looking at typing certain functions with Iterable. This is collections.abc.Iterable for 3.9+ but typing.Iterable for 3.7-3.8. Looking at https://pypistats.org/packages/jsonpickle, we only get ~60k downloads/day from 3.7 and 3.8 combined out of ~600k downloads on a normal weekday. That's roughly 10%, and 3.7 and 3.8 are EOL now too. Would it be reasonable to drop 3.7 and 3.8 support come 5.0.0?
3. pytest-monkeytype seems to be a nice tool for generating typing information, but I noticed that our pytest suite only covered around half of the functions in the util module. While I expect it to cover almost every function in other modules, it would still be nice to reach 100% coverage for the code we have left. I think I can try increasing coverage through the test suite over the summer (I'll have a full-time internship so no promises, but I was able to do a decent bit last summer).
4. The typing output by pytest-monkeytype seems to have a decent number of typing stubs that are fairly narrow because the pytest tests only cover 1-2 different input types for some functions. In addition to writing new tests that cover different input types, does the aetheris fuzzing module have any integrations with monkeytype or another automatic typer so that we can test different input types? If so, can we proceed by seeing which input types don't produce an error and adjust typing information accordingly?

What are y'all's thoughts on taking these actions, especially @davvid (and @DaveLak on point 4)?

Theelx:
This is sort of an omnibus PR ("big beautiful bill" --> "prodigious proposed PR"), with adding type hints to the public API, dropping support for python 3.7 and 3.8 and adding it for 3.13, moving some util functions to the private API so we can change them without worrying about backwards compatibility, and removing some deprecated functions. That's why there's a large amount of churn. I considered splitting each of the above updates into separate PRs but the function removal needed to be done before type hints, and dropping support for old python versions fit well with the new mypy addition to the testing matrix. **This is not everything that will be necessary for 5.0.0. so we can't tag v5 yet.**

You'll see that the commit times are almost all from 3 days ago, that's because I messed up some git merge/cherry-pick operation across branches and it reset the commit times. I've spent something like ~20 hours over the past week alone on this, so I'm going to have to take a break for a few days, then I'll move on to the rest of the 5.0.0 milestones.

Closes #441, #554, #561

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,7 +37,7 @@ jobs:
           restore-keys: |
             ${{ matrix.os }}-${{ matrix.python-version }}-v1-
 
-      - name: Install dependencies
+      - name: Install tox dependencies
         run: |
           python -m pip install -U pip
           python -m pip install -U "tox>=4"
@@ -51,3 +51,11 @@ jobs:
       - name: Tox tests
         run: |
           tox run
+
+      - name: Install mypy dependencies
+        run: |
+          python -m pip install -U .[testing]
+
+      - name: Mypy tests
+        run: |
+          mypy ./jsonpickle
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
         os: [ubuntu-latest]
 
     steps:
@@ -43,7 +43,7 @@ jobs:
           python -m pip install -U "tox>=4"
 
       - name: Install gmpy2 dependencies
-        if: matrix.python-version == '3.11' || matrix.python-version == '3.12'
+        if: matrix.python-version == '3.11' || matrix.python-version == '3.12' || matrix.python-version == '3.13'
         run: |
           sudo apt-get -y update
           sudo apt-get install -y libmpfr-dev libmpc-dev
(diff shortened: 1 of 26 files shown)


davvid (approved):
Very nicely done. Thanks for taking this on, this looks great.

PR #565, opened 2025-06-08
Fix handling of Exception subclasses with keywork-only arguments

### LINKED ISSUE #564: Cannot decode subclass of exception with keyword parameter
I am using jsonpickle v4.1.1.

I am likely doing something dumb, but I am getting an error when trying to decode into a subclass of Exception with one keyword only argument. I can minimally reproduce this as follows:

```py
import jsonpickle

class Foo(Exception):
    def __init__(self, a, *, b):
        self.a = a
        self.b = b

>>> jsonpickle.encode(Foo("a", b="b"))
'{"py/reduce": [{"py/type": "__main__.Foo"}, {"py/tuple": ["a"]}, {"a": "a", "b": "b"}]}'

>>> jsonpickle.decode(jsonpickle.encode(Foo("a", b="b")))
  File "/Users/dfarr/Developer/resonatehq-examples/deep-research/.venv/lib/python3.12/site-packages/jsonpickle/unpickler.py", line 103, in decode
    return context.restore(data, reset=reset, classes=classes)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dfarr/Developer/resonatehq-examples/deep-research/.venv/lib/python3.12/site-packages/jsonpickle/unpickler.py", line 420, in restore
    value = self._restore(obj)
            ^^^^^^^^^^^^^^^^^^
  File "/Users/dfarr/Developer/resonatehq-examples/deep-research/.venv/lib/python3.12/site-packages/jsonpickle/unpickler.py", line 402, in _restore
    return restore(obj)
           ^^^^^^^^^^^^
  File "/Users/dfarr/Developer/resonatehq-examples/deep-research/.venv/lib/python3.12/site-packages/jsonpickle/unpickler.py", line 555, in _restore_reduce
    stage1 = f(*args)
             ^^^^^^^^
TypeError: Foo.__init__() missing 1 required keyword-only argument: 'b'
```

I also get a similar error when Foo is defined with two positional arguments and I instantiate as `Foo("a", b="b")`. However, I do not get this error if I instantiate with two positional args in this case `Foo("a", "b")`

Theelx:
Fixes #564.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,13 +4,14 @@ v5.0.0
       You can re-enable yaml support using ``jsonpickle.ext.yaml.register()``.
       (#550) (+551)
     * **Breaking Change**: Support for CPython 3.7 and 3.8 was dropped, and support for
-      CPython 3.13 was added. (#561)
+      CPython 3.13 was added. (#561) (+563)
     * **Breaking Change**: Changed ``is_*`` functions in jsonpickle.util to the private API
-      rather than the public one by adding a single underscore to their names.
+      rather than the public one by adding a single underscore to their names. (+563)
     * **Breaking Change**: Removed the deprecated functions deprecated in 4.1.0 with commit
       59dda25. Additionally, ``util.is_sequence`` was removed due to having only one caller,
-      ``util.is_sequence_subclass``.
-    * Mypy-compatible typing has been added to the entire jsonpickle public API! (#561)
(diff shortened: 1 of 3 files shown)


davvid (approved):
It's really nice that this fix ended up being so simple.

PR #566, opened 2025-08-01
Fix type hint packaging with py.typed marker

Kakadus:
Hi, cool project!

I noticed that type hints were added recently. However, a `py.typed` marker is missing, so mypy won't pick up the inline type hints. Also see https://typing.python.org/en/latest/spec/distributing.html#packaging-type-information.

Before:
```py
error: Skipping analyzing "jsonpickle": module is installed, but missing library stubs or py.typed marker  [import]
note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
error: Class cannot subclass "BaseHandler" (has type "Any")  [misc]
```

After:
```py
Success: no issues found in 1 source file
```

Thanks :)


Theelx:
Looks great, thanks! I'll merge it once the tests complete.

PR #571, opened 2025-11-09
fix: change restore parameter from obj to data

Atry:
This change makes the signature of BaseHandler consistent with other handlers.

### DIFF
--- a/jsonpickle/handlers.py
+++ b/jsonpickle/handlers.py
@@ -124,7 +124,7 @@ class BaseHandler:
         """
         raise NotImplementedError('You must implement flatten() in %s' % self.__class__)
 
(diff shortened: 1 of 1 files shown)


Theelx:
Thank you for this! I would like to add though, I noticed the issues you created seemed AI-generated or at least AI-assisted. If this PR/bug was found with the help of AI, could you add an Assisted-by footer to your commit as suggested in https://docs.fedoraproject.org/en-US/council/policy/ai-contribution-policy/? I recently came across this and I think it's a good idea for transparency. If this wasn't found or made with AI, I'd be happy to merge it now, otherwise please amend the commit footer before I merge it.

PR #575, opened 2025-11-19
Add support for importable names of `types` module in `importable_name` function

### LINKED ISSUE #574: `types.XxxType` cannot round-trip
```
$ python
Python 3.13.8 (main, Oct  7 2025, 12:01:51) [GCC 14.3.0] on linux
Ctrl click to launch VS Code Native REPL
>>> import jsonpickle
>>> import types
>>> jsonpickle.decode(jsonpickle.encode(types.CoroutineType))
{'py/type': 'builtins.coroutine'}
>>> jsonpickle.__version__
'4.1.1'
```

Atry:
This PR fixes #574

### DIFF
--- a/jsonpickle/util.py
+++ b/jsonpickle/util.py
@@ -69,7 +69,11 @@ NON_CLASS_TYPES: set = {
     tuple,
     bytes,
 } | PRIMITIVES
-
+_TYPES_IMPORTABLE_NAMES = {
+    getattr(types, name): f"types.{name}"
+    for name in types.__all__
+    if name.endswith("Type")
+}
 
 def _is_type(obj: Any) -> bool:
     """Returns True is obj is a reference to a type.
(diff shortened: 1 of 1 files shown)


Theelx:
I personally am fine with the dynamic type getting. Our test suite is fairly extensive so if there are issues with other types we'll probably see it in the tests. Plus, given the pace at which Python has been iterating and improving since around 3.11, it'd be nice not to have to manually update the list every time a new Python version comes out. Could you fix the failing doctest and add a roundtrip test in tests/jsonpickle_test.py though?

Atry:
Will do

jsonpickle/util.py
```diff
@@ -481,8 +486,14 @@ def importable_name(cls: Union[Type, Callable[..., Any]]) -> str:
     True
     >>> importable_name(AttributeError) == 'builtins.AttributeError'
     True
+    >>> importable_name(type(argparse.ArgumentParser().add_argument)) == 'types.MethodType'
+    True
 
     """
+    types_importable_name = _TYPES_IMPORTABLE_NAMES.get(id(cls))
```
Theelx: Sorry if this is a stupid question (I just got back home from the airport and am super jet lagged) but if two objects of different types have the [same id because of non-overlapping lifetimes](https://docs.python.org/3/library/functions.html#id), could this break, especially if someone is dynamically adding types?

Atry: Do you think if it makes sense to use the type as the key?

Theelx (approved):
Looks great, thanks!

Issue #579, opened 2025-12-02
Deallocation Of Memory After Decode


danbodily:
See the below script.   I'm trying to control when memory is freed after a JSON decode.   I'm finding that I cannot remove references that are setup in jsonpickle during decoding.  Is there a "right" way to do this?  I found one way was to have the pickle load actually occur in a child process using the multiprocessing library but that is slow and seems overly complex...
```
import jsonpickle
import jsonpickle.ext.numpy as jsonpickle_numpy
jsonpickle_numpy.register_handlers()
jsonpickle.set_encoder_options('simplejson', indent=4)
import numpy as np


class MyClass(object):
    def __init__(self):
        self.a = np.array([0,0,1])

    def __del__(self):
        print("DELETED", flush = True)

m = MyClass()
g = jsonpickle.encode(m)
m = None
# DELETE IS CALLED HERE

m2 = jsonpickle.decode(g)
m2 = None
# WHY IS DELETE IS NOT CALLED HERE??

print("END OF SCRIPT", flush = True)
```
Script yields output:
DELETED
END OF SCRIPT
DELETED

Expected:
DELETED
DELETED
END OF SCRIPT

PR #580, opened 2025-12-07
unpickler: avoid holding on to external references

davvid:
Even though the unpickler goes out of scope immediately after decode() completes, the Python garbage collector is not able to determine that garbage collection can occur despite the fact that no references are being held. The observed behavior is that Python is not deleting the ephemeral unpickler instance.

Make this easier for the GC to handle by explicitly resetting the ephemeral Unpickler instance at the end of decode(). This guarantees that no external references are being held and allows them to be cleaned up in a slightly more predictable fashion.

Closes: #579

### DIFF
--- a/jsonpickle/unpickler.py
+++ b/jsonpickle/unpickler.py
@@ -111,6 +111,7 @@ def decode(
         )
 
     backend = backend or json
+    is_ephemeral_context = context is None
     context = context or Unpickler(
         keys=keys,
         backend=backend,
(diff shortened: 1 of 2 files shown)


davvid:
The most recent push added a test to ensure that this behaves as expected. I was only able to make the test fail when using numpy (as reported in #579) so I ended up adding the test to the numpy extension tests in `numpy_test.py`.

Theelx (approved):
This is an surprisingly simple fix, thanks for this! I added my approval, but can you add a CHANGES.rst entry before merging it?

PR #603, opened 2026-06-03
Modernize type annotations

jorenham:

In order for us to merge your PR, please confirm that all of these boxes are true when applied to this PR and check them if so:

- [ ] If this PR changes any non-documentation portion of the codebase, I have updated the ``CHANGES.rst`` file for this change.
- [ ] If this PR fixes a bug or adds a feature, I have added appropriate tests.
- [x] I have run the tests with ``pytest`` and run formatting with ``black``.
- [ ] If generative AI of any kind was used in creating this PR, I have followed the these guideliens:
   - [ ] I ensured that this PR is not entirely written by any generative AI model.
   - [ ] I ensured that any portions of the code written by any generative AI model are free from copyrighted code.
   - [ ] I ensured that I included a Co-authored-by or Assisted-by tag in the footer containing the name of all generative AI models used.
   - [ ] I ensured that all code in this PR submitted verbatim from generative AI output is limited to creating/fixing tests/documentation and/or fixing bugs in extensions, and does not impact the core code.

---

I noticed some deprecated/legacy (pre Python 3.10) typing syntax, so I figured I'd modernize it a bit. 

Everything passes with the latest mypy 2.1.0. There was no upper verison bound for mypy in the dev requirements, so for consistency I also upgraded the pre-commit mypy version to the latest version.

I'm not sure whether I should update `CHANGES.rst` because these changes are all non-functional.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,7 +19,7 @@ repos:
       - --profile=black
 
 - repo: https://github.com/pre-commit/mirrors-mypy
-  rev: v1.19.1
+  rev: v2.1.0
   hooks:
   - id: mypy
     types: [python]
--- a/jsonpickle/backend.py
+++ b/jsonpickle/backend.py
@@ -1,5 +1,5 @@
 from types import ModuleType
-from typing import Any, Dict, Optional, Type, Union
+from typing import Any
 
 
 class JSONBackend:
@@ -20,7 +20,7 @@ class JSONBackend:
         raise AssertionError("jsonpickle could not load any json modules")
 
     def encode(
-        self, obj: Any, indent: Optional[int] = None, separators: Optional[Any] = None
+        self, obj: Any, indent: int | None = None, separators: Any | None = None
     ) -> str:
         """
         Attempt to encode an object into JSON.
@@ -44,7 +44,7 @@ class JSONBackend:
                 if idx == len(self._backend_names) - 1:
                     raise e
 
-    # def dumps
+    # def dumpsscipy
     dumps = encode
 
     def decode(self, string: str) -> Any:
@@ -126,7 +126,7 @@ class JSONBackend:
         self._fallthrough = enable
 
     def _store(
-        self, dct: Dict[str, Any], backend: str, obj: ModuleType, name: str
+        self, dct: dict[str, Any], backend: str, obj: ModuleType, name: str
     ) -> bool:
         try:
             dct[backend] = getattr(obj, name)
@@ -140,7 +140,7 @@ class JSONBackend:
         name: str,
         dumps: str = "dumps",
         loads: str = "loads",
-        loads_exc: Union[str, Type[Exception]] = ValueError,
+        loads_exc: str | type[Exception] = ValueError,
     ) -> bool:
         """Load a JSON backend by name.
 
@@ -214,8 +214,8 @@ class JSONBackend:
         self,
         name: str,
         obj: Any,
-        indent: Optional[int] = None,
-        separators: Optional[str] = None,
+        indent: int | None = None,
+        separators: str | None = None,
     ) -> str:
         optargs, optkwargs = self._encoder_options.get(name, ([], {}))
         encoder_kwargs = optkwargs.copy()
--- a/jsonpickle/backend.py
+++ b/jsonpickle/backend.py
@@ -44,7 +44,7 @@ class JSONBackend:
                 if idx == len(self._backend_names) - 1:
                     raise e
 
-    # def dumpsscipy
+    # def dumps
     dumps = encode
 
     def decode(self, string: str) -> Any:
--- a/jsonpickle/errors.py
+++ b/jsonpickle/errors.py
@@ -2,9 +2,9 @@
 Stores custom jsonpickle errors.
 """
 
-from typing import Any, Dict
+from typing import Any
 
 
 class ClassNotFoundError(BaseException):
-    def __init__(*args, **kwargs: Dict[str, Any]):
+    def __init__(*args: Any, **kwargs: Any):
         pass
--- a/jsonpickle/ext/numpy.py
+++ b/jsonpickle/ext/numpy.py
@@ -6,21 +6,24 @@ import sys
 import warnings
 import zlib
 from types import ModuleType
-from typing import Any, NoReturn, cast
+from typing import TYPE_CHECKING, Any, NoReturn, cast
 
 import numpy as np
 
 # do the annotations import so python doesn't complain about numpy type hints missing if numpy isn't installed
 
 
-_np_version: tuple[int, ...] = tuple(int(x) for x in np.__version__.split(".")[:3])
-# numpy.typing was introduced in 1.20.0
-if _np_version >= (1, 20, 0):
+if TYPE_CHECKING:
     from numpy.typing import ArrayLike, DTypeLike, NDArray
 else:
-    NDArray = Any  # type: ignore[assignment,misc]
-    ArrayLike = Any  # type: ignore[assignment,misc]
-    DTypeLike = Any  # type: ignore[assignment,misc]
+    _np_version: tuple[int, ...] = tuple(int(x) for x in np.__version__.split(".")[:3])
+    # numpy.typing was introduced in 1.20.0
+    if _np_version >= (1, 20, 0):
+        from numpy.typing import ArrayLike, DTypeLike, NDArray
+    else:
+        NDArray = Any
+        ArrayLike = Any
+        DTypeLike = Any
 
 
 from ..handlers import BaseHandler, register, unregister
@@ -49,7 +52,7 @@ class NumpyBaseHandler(BaseHandler):
                 dtype = dtype[len(prefix) : -1]
             data["dtype"] = dtype
 
-    def restore_dtype(self, data: dict[str, Any]) -> np.dtype:  # type: ignore[type-arg]
+    def restore_dtype(self, data: dict[str, Any]) -> np.dtype[Any]:
         dtype = data["dtype"]
         if dtype.startswith(("{", "[")):
             dtype = ast.literal_eval(dtype)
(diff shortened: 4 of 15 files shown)


jsonpickle/ext/numpy.py
```diff
@@ -6,21 +6,24 @@
...
+if TYPE_CHECKING:
     from numpy.typing import ArrayLike, DTypeLike, NDArray
 else:
-    NDArray = Any  # type: ignore[assignment,misc]
-    ArrayLike = Any  # type: ignore[assignment,misc]
-    DTypeLike = Any  # type: ignore[assignment,misc]
+    _np_version: tuple[int, ...] = tuple(int(x) for x in np.__version__.split(".")[:3])
+    # numpy.typing was introduced in 1.20.0
+    if _np_version >= (1, 20, 0):
+        from numpy.typing import ArrayLike, DTypeLike, NDArray
```
davvid: This part is duplicated now, and the duplicate section above dropped the version check. It seems like this should also remove the imports above so that we only import inside the version-check blocks.

.pre-commit-config.yaml
```diff
@@ -19,10 +19,10 @@ repos:
...
-  rev: v1.19.1
+  rev: v2.1.0
   hooks:
   - id: mypy
     types: [python]
     args: [--strict, --config-file, mypy.ini, jsonpickle]
     pass_filenames: false
-    language: system
\ No newline at end of file
+    language: system
```
davvid, line 28: Please keep these changes in a separate commit if possible.

If they have to go together then that's okay.. just noting because this stood out.

davvid (changes requested):
Nicely done, looks good. I just had a few minor notes. I'm okay to merge once these are addressed.

jsonpickle/ext/numpy.py
jorenham: That was intented, because this part is needed for *runtime* type checkers (beartype, typeguard, etc.), whereas the `TYPE_CHECKING` bit is necessary for *static* type checkers (mypy, pyright, pyrefly, etc).

jorenham: But I agree it looks awkward. Is dropping support for numpy <1.20 an option maybe?

Theelx:
I can check over these tonight (ET), can we wait to merge for at least a few hours?

jsonpickle/ext/numpy.py
Theelx: Yeah, I think dropping numpy <1.20 is viable. I think I chose support for numpy <1.20 back when we supported Python 3.6, since 1.19.x was the last version that supported 3.6. Now that we support no lower than 3.10, this should be fine.

jsonpickle/backend.py
```diff
@@ -44,7 +44,7 @@ def encode(
                 if idx == len(self._backend_names) - 1:
                     raise e
 
-    # def dumps
+    # def dumpsscipy
```
Theelx: Was this a copy-paste or find-replace error?

jsonpickle/errors.py
```diff
@@ -2,9 +2,9 @@
 Stores custom jsonpickle errors.
 """
 
-from typing import Any, Dict
+from typing import Any
 
 
 class ClassNotFoundError(BaseException):
-    def __init__(*args, **kwargs: Dict[str, Any]):
```
Theelx, line 9: Can I ask why you removed the Dict[str, ...] part? I was under the impression that kwargs could only be a dict of strings to something else (at least, barring some C-extension trickery), and I feel that if that's a correct assumption then the more explicit, the better.

jsonpickle/pickler.py
```diff
@@ -22,18 +23,18 @@ def encode(
...
     keys: bool = False,
-    max_depth: Optional[int] = None,
+    max_depth: int | None = None,
     reset: bool = True,
-    backend: Optional[JSONBackend] = None,
+    backend: JSONBackend | None = None,
     warn: bool = False,
-    context: Optional["Pickler"] = None,
-    max_iter: Optional[int] = None,
+    context: "Pickler | None" = None,
```
Theelx, line 30: Should the "| None" be going inside the quotes?

jsonpickle/pickler.py
```diff
@@ -416,11 +417,12 @@ def _flatten_obj(self, obj: Any) -> Any:
 
         try:
             in_cycle = _in_cycle(obj, self._objs, max_reached, self.make_refs)
+            flatten_func: Callable[[Any], str] | None
```
Theelx, line 420: Does this need an "if TYPE_CHECKING" guard before it? I'm fine with whatever solution ends up with it not having to run at runtime when the user to actually using the library.

jsonpickle/pickler.py
```diff
@@ -456,8 +458,8 @@ def _getstate(self, obj: Any, data: Dict[str, Any]) -> Dict[str, Any]:
         return data
 
     def _flatten_key_value_pair(
-        self, k: Any, v: Any, data: Dict[Union[str, Any], Any]
-    ) -> Dict[Union[str, Any], Any]:
+        self, k: Any, v: Any, data: dict[str | Any, Any]
```
Theelx, line 461: I just realized that I think this doesn't make sense the way I initially wrote it. The `str | any` bit doesn't make much sense because that could just be `Any`. However, this may be useful for being explicit. Thoughts?

jsonpickle/pickler.py
jorenham: Yes; `"Pickler" | None` will raise an error.

jsonpickle/pickler.py
jorenham: `str | Any` isn't the same as `Any`; it's a type that's either `str` or something else. So you can read it as `str | T` with `T` free.

jsonpickle/backend.py
jorenham: oh yea woops; not sure what went wrong here

jsonpickle/errors.py
jorenham: `kwargs` itself can indeed only be a dict; but if you write `**kwargs: dict[str, Any]`, then **each** kwarg only accepts a `dict[str, Any]`. So it's the kwarg *value types* that are annotated, not the `kwargs` type itself.

jsonpickle/pickler.py
jorenham: There's no need for that; `Callable[[Any], str]` evaluates just fine at runtime on Python 3.10+.

Also, this can help tell runtime type-checker (beartype, typeguard, etc) tell what's going on.

jsonpickle/ext/numpy.py
```diff
@@ -6,40 +6,34 @@
...
+
+
 native_byteorder: str = "<" if sys.byteorder == "little" else ">"
 
 
-# considered typing arr as ArrayLike but then we get a mypy error with no attribute dtype
-def get_byteorder(arr: ArrayLike) -> str:
+def get_byteorder(arr: _HasDType) -> str:
     """translate equals sign to native order"""
-    byteorder = arr.dtype.byteorder  # type: ignore[union-attr]
```
jorenham, line 37: `ArrayLike` also accepts types such as `float` and `str`, which have no `.dtype` attribute, so mypy was right to complain here.

jsonpickle/ext/numpy.py
```diff
@@ -6,40 +6,34 @@
...
 from ..handlers import BaseHandler, register, unregister
 from ..util import b64decode, b64encode, importable_name, loadclass
 
 __all__ = ["register_handlers", "unregister_handlers"]
 
+
+# matches `np.ndarray`, `np.generic`, and anything else with a `.dtype` attribute
+class _HasDType(Protocol):
+    @property
+    def dtype(self) -> np.dtype[Any]: ...
```
jorenham, line 23: This might seem hacky; but we use this "structural typing" approach internally in NumPy all the time; so it's battle tested.

pyproject.toml
```diff
@@ -56,7 +56,7 @@ testing = [
     "feedparser",
     "gmpy2",
     "isort",
-    "numpy",
+    "numpy>=1.20",
```
jorenham: NumPy 1.20.3 doesn't support Python 3.10 (3.9 at most). NumPy 1.21 is the first that supports Python 3.10.

- https://pypi.org/project/numpy/1.20.3/
- https://pypi.org/project/numpy/1.21.6/

Maybe we should increase this to `>=1.21 `?

jorenham: Removing the bound here is also an option I guess, because it'll implicitly be `>=1.21` either way. But I guess there's also something to be said for being explicit.

jsonpickle/pickler.py
Theelx: Got it. Are there any potential ramifications (such as confusion) from users or tools that use e.g. inline hints in editors like VSCode thinking that the class is named something that seems invalid like a literal "Pickler | None"?

jsonpickle/errors.py
Theelx: Got it, thanks for clearing that up!

jsonpickle/ext/numpy.py
Theelx: Interesting. Thanks for bringing over that knowledge!

pyproject.toml
Theelx: I'd appreciate an explicit >=1.21.

jsonpickle/pickler.py
jorenham: Hm no not really. Even runtime type-checkers like bearype are usually able to deal with stringified annotations like this one. 

Many people use `from __future__ import annotations`, which stringifies every annotation, and that also doesn't lead to any issues in IDEs and related tooling as far as I'm aware.

Theelx:
Last change from me: could you add (+603) to this line in CHANGES.rst? Additionally, feel free to add yourself to the end of the authors list in docs/index.rst, as this is a significant contribution. If you'd rather not be on it, that's fine too; it's up to you.
> Mypy-compatible typing has been added to the entire jsonpickle public API! (#561) (+563)

PR #606, opened 2026-06-17
Preserve bytearray contents through encode/decode

gaoflow:
### Problem

A non-empty `bytearray` does not survive a roundtrip; it is silently emptied:

```python
import jsonpickle
data = bytearray(b"\x00\x01\xff hello")
jsonpickle.decode(jsonpickle.encode(data))   # -> bytearray(b'')
```

`encode` produces `{"py/object": "builtins.bytearray"}` (no state), and `decode`
returns an empty `bytearray`. This is silent data loss rather than an error.

### Cause

In `Pickler._flatten_obj` the only bytestring special-case is
`if type(obj) is bytes:`. A `bytearray` is not a `bytes` instance, and because it
is list-like, `util._is_reducible` returns `False`, so it never reaches the
`py/reduce` path either. It falls through to the generic object flatten, which
has no meaningful state to serialize for a `bytearray`, so the contents are
dropped.

### Fix

Encode a `bytearray` like a bytestring under a new `py/bytearray` tag, reusing
the existing `_flatten_bytestring` helper so it honors `use_base85`, and restore
it back into a `bytearray`. This mirrors how `bytes` already roundtrips via
`py/b64` (the documented oracle: the mutable sibling type should preserve its
data just as `bytes` does).

Encoded form: `{"py/bytearray": {"py/b64": "AAH/IGhlbGxv"}}` (or `py/b85` under
`use_base85=True`).

Exact-type checks (`type(obj) is bytearray`) are used so `bytearray` subclasses
keep their current behavior. The new tag is added to `RESERVED`, and decoding of
existing payloads is unaffected (the tag is additive).

### Tests

Added `test_bytearray_roundtrip`, `test_bytearray_empty_roundtrip`, and
`test_bytearray_nested_roundtrip` to `tests/jsonpickle_test.py`, covering content
preservation, the empty case, and nesting inside a container. Verified red before
the fix and green after; the rest of the suite is unaffected.

---

*Disclosure: I investigated and prepared this patch with the help of an AI
assistant, working under my direction. I reviewed the encoder/decoder paths,
confirmed the root cause, and verified the tests (red then green) myself.*

### DIFF
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -389,14 +389,19 @@ class Pickler:
         #########################################
         # if obj is nonrecursive return immediately
         # for performance reasons we don't want to do recursive checks
-        if type(obj) is bytes:
+        typeof_obj = type(obj)
+        if typeof_obj is bytes:
             return self._flatten_bytestring(obj)
 
         # Decimal is a primitive when use_decimal is True
-        if type(obj) in (str, bool, int, float, type(None)) or (
+        if typeof_obj in (str, bool, int, float, type(None)) or (
             self._use_decimal and isinstance(obj, decimal.Decimal)
         ):
             return obj
+
+        # bytearray is list-like, so it is neither reducible nor atomic.
+        if typeof_obj is bytearray:
+            return {tags.BYTEARRAY: self._flatten_bytestring(bytes(obj))}
         #########################################
 
         self._push()
--- a/jsonpickle/tags.py
+++ b/jsonpickle/tags.py
@@ -8,6 +8,7 @@ that need to be specially handled.
 """
 
 BYTES: str = "py/bytes"
+BYTEARRAY: str = "py/bytearray"
 B64: str = "py/b64"
 B85: str = "py/b85"
 FUNCTION: str = "py/function"
@@ -34,6 +35,7 @@ TYPE: str = "py/type"
 # All reserved tag names
 RESERVED: set[str] = {
     BYTES,
+    BYTEARRAY,
     FUNCTION,
     ID,
     INITARGS,
--- a/jsonpickle/tags.py
+++ b/jsonpickle/tags.py
@@ -8,7 +8,7 @@ that need to be specially handled.
 """
 
 BYTES: str = "py/bytes"
-BYTEARRAY: str = "py/bytearray"
+BYTEARRAY: str = "py/bytea"
 B64: str = "py/b64"
 B85: str = "py/b85"
 FUNCTION: str = "py/function"
--- a/jsonpickle/unpickler.py
+++ b/jsonpickle/unpickler.py
@@ -438,6 +438,14 @@ class Unpickler:
         except (AttributeError, UnicodeEncodeError):
             return b""
 
+    def _restore_bytearray(self, obj: dict[str, Any]) -> bytearray:
+        payload = obj[tags.BYTEARRAY]
+        if tags.B85 in payload:
+            data = self._restore_base85(payload)
+        else:
+            data = self._restore_base64(payload)
+        return bytearray(data)
+
     def _refname(self) -> str:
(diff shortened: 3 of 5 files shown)


Theelx:
Thank you for your contribution, and I apologize that I let this slip under my radar for 2 weeks! I must have accidentally marked this notification as read when I wasn't thinking properly. This fix looks all good for me. Could you just retroactively add an Assisted-By footer with the name and version of the AI model that you used to each commit where it was used? This will help significantly in case we ever have to go back and identify AI-assisted commits en masse for legal/copyright reasons.

Since the change to the core pickler.py and unpickler.py files is so small and trivial, since I understand it perfectly fine myself (so I can take on the maintenance burden), and since you've clearly done manual review, I think I'm fine with letting the AI usage on this change into the codebase.

jsonpickle/pickler.py
```diff
@@ -392,6 +392,12 @@ def _flatten_impl(self, obj: Any) -> Any:
         if type(obj) is bytes:
             return self._flatten_bytestring(obj)
 
+        if type(obj) is bytearray:
+            # bytearray is not reducible (it is list-like) and is not bytes,
+            # so encode its contents like a bytestring and tag it so that it
+            # is restored as a bytearray rather than losing its data.
+            return {tags.BYTEARRAY: self._flatten_bytestring(bytes(obj))}
+
```
davvid: Given the frequency of the atomic types below (str, bool, int, float, etc) I think this check would best be done immediately after that check below rather than here. It'll at least save us a whole conditional block in common cases where bytearray isn't being processed.


Slightly related, but we should probably cache `typeof_obj = type(obj)` early in this function since we use it so often. We're paying for a function call into `type(..)` in each expression, which isn't really that cheap in python.

jsonpickle/unpickler.py
```diff
@@ -439,6 +439,14 @@ def _restore_base85(self, obj: dict[str, Any]) -> bytes:
         except (AttributeError, UnicodeEncodeError):
             return b""
 
+    def _restore_bytearray(self, obj: dict[str, Any]) -> bytearray:
+        payload = obj[tags.BYTEARRAY]
+        if isinstance(payload, dict) and tags.B85 in payload:
```
davvid: The isinstance() check is misleading here, because the else: side of it happily assumes that it's true as well.

It's almost better to remove it. The code can just assume that the data is good. If it's not it'll rightfully throw an exception.

davvid (changes requested):
Nicely done, this looks pretty close to being good to merge to me. Please amend your commits to address these notes and we'll re-review. Thanks

binggao1230:
Moved the bytearray branch below the atomic-type check and cached `typeof_obj = type(obj)` for the three type tests in `_flatten_impl`; the reorder is safe because the only check now sitting in front of it can't match an exact bytearray, and diffing encode output over ~2600 object/option combinations before and after showed no change. Dropped the `isinstance` in `_restore_bytearray` — `tags.B85 in payload` selects the same branch for everything jsonpickle emits, and a malformed payload still raises TypeError exactly as before. Added b64/b85 flatten and restore tests plus a base85 round-trip, which takes the suite from 405 to 410 passing with nothing else moving. Also went back and added the Assisted-by trailers @Theelx asked for to both commits.

jsonpickle/tags.py
```diff
@@ -8,6 +8,7 @@
 """
 
 BYTES: str = "py/bytes"
+BYTEARRAY: str = "py/bytearray"
```
Theelx: I think this is my last comment before merging: can this actually be "py/bytea"? I think bytea is still clear to a human while taking up fewer characters.

Theelx (changes requested):
Looks good apart from this last minor change!

binggao1230:
Done — `py/bytea` (`3353c22`), rebased onto main.

PR #607, opened 2026-06-24
Fix RegexHandler to preserve flags in encode/decode round-trip

gaoflow:
## Summary

`jsonpickle.encode(re.compile(pattern, re.IGNORECASE | re.MULTILINE))` followed by `decode()` silently drops all non-default flags, returning a pattern with only `re.UNICODE` set.

**Root cause:** `RegexHandler.flatten` stored only the pattern string; `restore` called `re.compile(data["pattern"])` with no flags argument.

**Fix:** Store `obj.flags` in the flattened dict and pass it to `re.compile()` on restore.

```python
# Before fix
import jsonpickle, re
pat = re.compile(r'\d+', re.IGNORECASE | re.MULTILINE)
decoded = jsonpickle.decode(jsonpickle.encode(pat))
assert decoded.flags == pat.flags  # FAILS: decoded.flags == 32, expected 42

# After fix
decoded = jsonpickle.decode(jsonpickle.encode(pat))
assert decoded.flags == pat.flags  # PASSES
```

## Changes

- `jsonpickle/handlers.py`: `RegexHandler.flatten` now stores `flags`; `restore` passes it to `re.compile`.
- `tests/object_test.py`: new test `test_compiled_regex_flags_roundtrip` covering `re.IGNORECASE | re.MULTILINE`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -42,6 +42,8 @@ v5.0.0
       The decorative ``values`` emitted for a referenced array no longer consume a
       ``py/id`` slot, keeping the encode and decode reference counters in sync.
       (#449) (+612)
+    * Fix bug where compiled regex objects lost their flags (e.g. ``re.IGNORECASE``)
+      when roundtripped through encode/decode. (+607)
 
 v4.1.2
 ======
--- a/jsonpickle/handlers.py
+++ b/jsonpickle/handlers.py
@@ -242,10 +242,11 @@ class RegexHandler(BaseHandler):
 
     def flatten(self, obj: re.Pattern[str], data: dict[str, Any]) -> HandlerReturn:
         data["pattern"] = obj.pattern
+        data["flags"] = obj.flags
         return data
 
     def restore(self, data: dict[str, Any]) -> re.Pattern[str]:
-        return re.compile(data["pattern"])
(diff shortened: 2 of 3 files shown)


Theelx:
Thank you for this! This looks all good for me. I only have two minor requests. First, if you used an AI model then could you retroactively add an Assisted-By footer with the name and version of the AI model that you used to each commit where it was used? If you didn't use an AI model, please simply state that in a comment here so we have a record. Second, could you add an entry to CHANGES.rst for this change? After these, I'd be happy to merge!

binggao1230:
Both commits now carry an `Assisted-by` trailer: Claude Opus 4 for the June fix, Claude Opus 5 for the new `CHANGES.rst` entry. I rebased onto current main so the changelog line lands after `(+612)` instead of conflicting with it; the suite goes 401 -> 402 passed with nothing else moving.

One thing to flag before you merge — the `handlers.py` change is core code rather than tests/docs, so I can't claim that condition from the template checklist. You waived it on #606; mentioning it since you hadn't here.

Theelx:
This is such a tiny change that I'm fine with it touching core code; 2 lines is incredibly small. I'll merge this once tests pass.

Theelx:
The tests are failing because of a ruff change (not anything you did). Merging.

PR #609, opened 2026-07-01
Encode __reduce__-only objects instead of null when unpicklable=False

### LINKED ISSUE #444: datetime.timedelta not picklable if unpicklable == False
```python
>>> import datetime
>>> import jsonpickle

>>> delta = datetime.timedelta(days=3)

>>> jsonpickle.encode(delta, unpicklable=True)
'{"py/reduce": [{"py/type": "datetime.timedelta"}, {"py/tuple": [3, 0, 0]}]}'

>>> jsonpickle.encode(delta, unpicklable=False)
'null'
```

But, for example, other "lossy" one-way encodings are supported with unpicklable `True` or `False`:
```python
>>> tpl = (3, 0, 0)
>>> jsonpickle.encode(tpl, unpicklable=True)
'{"py/tuple": [3, 0, 0]}'
>>> jsonpickle.encode(tpl, unpicklable=False)
'[3, 0, 0]'
```

apoorva-01:
Closes #444.

`encode(timedelta(days=3), unpicklable=False)` returns `null` because the `__reduce__`/`__reduce_ex__` handling in `_flatten_obj_instance` is gated behind `if self.unpicklable:`. Objects whose state is reachable only through `__reduce__`(no `__dict__`/`__slots__`/`__getstate__`) then produce nothing and fall through to the catchall `return None`.

This extracts the reduce call into a `_reduce()` helper (the `unpicklable=True`path is unchanged) and adds a fallback just before that `return None`: emit a lossy view of the reduce output — its state if present, else the constructor args. `timedelta` now gives `[3, 1, 0]`. The fallback only runs where the encoder was already returning `None`, so nothing that encodes today changes.

Open question for you: should `unpicklable=False` prefer `__reduce__` for *all* such objects (as the reporter suggested) rather than only as a last resort? That one wanted to wait for v4 back in 2023 — now that main is 5.0 it's viable, via a wider default or an opt-in flag. Happy to follow your call; kept this contained for now.

Tested via a regression test in `tests/jsonpickle_test.py` (fails pre-fix) and the full suite incl. the numpy/pandas extension tests (415 passed / 1 skipped), plus `ruff` and `isort`.

---

- [x] I have run the tests with ``pytest`` and run formatting (``ruff format``).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -32,6 +32,9 @@ v5.0.0
     * Updated outdated documentation and added more usage examples (+594)
     * ``isort`` is now an optional "testing" dependency in ``pyproject.toml``.
     * The ``garden check`` command was corrected. (+604)
+    * Fix bug where objects whose state is only available through ``__reduce__``
+      (such as ``datetime.timedelta``) were encoded as ``null`` when
+      ``unpicklable=False``. They now fall back to a lossy representation. (#444)
 
 v4.1.2
 ======
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -579,6 +579,22 @@ class Pickler:
 
         return data
 
+    def _reduce(self, obj: Any, has_reduce: bool, has_reduce_ex: bool) -> Any:
+        """Return the object's __reduce__/__reduce_ex__ output, or None.
+
+        Many builtin types raise TypeError from these; treat that as
+        "no reduce available" rather than letting it propagate.
+        """
+        try:
+            if has_reduce and not has_reduce_ex:
+                return obj.__reduce__()
+            if has_reduce_ex:
+                # we're implementing protocol 2
+                return obj.__reduce_ex__(2)
+        except TypeError:
+            pass
+        return None
+
     def _flatten_obj_instance(
         self, obj: Any
     ) -> dict[str, Any] | list[Any] | Any | None:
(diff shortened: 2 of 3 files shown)


Theelx:
Thank you for making this! I have no major objections to the content of this PR, but I do have two minor nitpicks around AI usage. First, could you just clarify the version of Claude that you used? I realize now that I forgot to include a request in the PR template for the version of any models used to be put in the commit footer, sorry for the confusion! Second, it appears that some of the code in this PR that touches the core pickler/unpickler code is changed inside a commit with AI assistance. In the case that you wrote the code touching the core bits and AI helped with tests, then please separate the commit into one with the core changes without the assisted-by footer and one with the other changes (like tests and the changelog) that does have the assisted-by footer. If AI wrote the code touching the core files, then please let us know so that davvid and I can discuss how we can find a way to merge your contributions in a responsible manner.

This is not a denial of your PR, as I do recognize the effort you put in. I'd be more than happy to merge it once we get feedback from @davvid and once these nitpicks are addressed!

apoorva-01:
It was Claude Opus 4.8.

Split it into two commits: the core `pickler.py` change on its own with no footer (I wrote that part myself), and the tests + `CHANGES.rst` in a second commit that carries the `Assisted-by: Claude Opus 4.8 (Anthropic)` footer.

Happy to wait for davvid, and to move things around if you'd like it done differently.

davvid:
Thanks! FWIW this looks good to me.

Theelx:
Sounds great, thank you again @apoorva-01!

PR #611, opened 2026-07-05
Escape reserved tag dict keys before picklability checks

jvdillon:
- [x] I have run the tests with ``pytest`` and run formatting with ``black``.

---

Escape reserved tag dict keys before picklability checks.

- Prevent silent loss of dict keys equal to reserved wire tags
- Move escape earlier to catch keys before _is_picklable rejects them
- Add regression tests for reserved-tag dict keys with keys=True

### DIFF
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -828,13 +828,24 @@ class Pickler:
         self, k: str, v: Any, data: dict[str, Any]
     ) -> dict[str, Any]:
         """Flatten string key/value pairs only."""
+        if self.keys and isinstance(k, str) and (
+            k.startswith(tags.JSON_KEY) or k in tags.RESERVED
+        ):
+            # Escape a data key that collides with the json:// escape prefix or
+            # with a reserved wire tag (e.g. "py/object", "py/id"). This must run
+            # BEFORE the _is_picklable guard below: _is_picklable rejects any
+            # name in tags.RESERVED, which is correct for object attributes but
+            # would silently DROP a legitimate user dict key here. Without the
+            # escape the key is lost on encode, and on decode _restore_from_dict
+            # skips reserved keys and the node can be misrouted (e.g. treated as
+            # a py/id).
+            data[self._escape_key(k)] = self._flatten(v)
+            return data
         if not util._is_picklable(k, v):
             return data
         if self.keys:
             if not isinstance(k, str):
                 return data
-            elif k.startswith(tags.JSON_KEY):
-                k = self._escape_key(k)
         else:
             if k is None:
                 k = "null"  # for compatibility with common json encoders
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -828,17 +828,9 @@ class Pickler:
         self, k: str, v: Any, data: dict[str, Any]
     ) -> dict[str, Any]:
         """Flatten string key/value pairs only."""
-        if self.keys and isinstance(k, str) and (
-            k.startswith(tags.JSON_KEY) or k in tags.RESERVED
-        ):
-            # Escape a data key that collides with the json:// escape prefix or
-            # with a reserved wire tag (e.g. "py/object", "py/id"). This must run
-            # BEFORE the _is_picklable guard below: _is_picklable rejects any
-            # name in tags.RESERVED, which is correct for object attributes but
-            # would silently DROP a legitimate user dict key here. Without the
-            # escape the key is lost on encode, and on decode _restore_from_dict
-            # skips reserved keys and the node can be misrouted (e.g. treated as
-            # a py/id).
+        if (k.startswith(tags.JSON_KEY) or k in tags.RESERVED) and self.keys:
+            # Escape data keys colliding with the json:// prefix or a reserved
+            # wire tag; must run before _is_picklable, which drops RESERVED keys.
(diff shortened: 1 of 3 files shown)


tests/jsonpickle_test.py
```diff
@@ -936,6 +936,28 @@ def test_string_key_requiring_escape_dict_keys_with_keys_enabled():
     assert unpickled[tags.JSON_KEY + "6"] == [1, 2]
 
 
+@pytest.mark.parametrize("reserved", sorted(tags.RESERVED))
```
Theelx, line 939: Why is ``sorted()`` called here on a single item?

tests/jsonpickle_test.py
```diff
@@ -936,6 +936,28 @@ def test_string_key_requiring_escape_dict_keys_with_keys_enabled():
...
+def test_reserved_tag_dict_keys_with_keys_enabled(reserved):
+    """A user dict key equal to a reserved wire tag round-trips verbatim.
+
+    Regression: keys such as "py/object" / "py/id" were silently dropped on
```
Theelx: I don't believe this description of the regression is necessary in the docstring, the CHANGES.rst entry that you'll need to add that then links to this PR should be plenty.

jsonpickle/pickler.py
```diff
@@ -828,13 +828,24 @@ def _flatten_string_key_value_pair(
         self, k: str, v: Any, data: dict[str, Any]
     ) -> dict[str, Any]:
         """Flatten string key/value pairs only."""
+        if self.keys and isinstance(k, str) and (
```
Theelx: self.keys will default to True in v5, which is when this will be released. Could you move the ``if self.keys`` check to the very end of the if-statement so that the more selective statements come first? Also, I don't believe the ``isinstance(k, str)`` check is necessary because this function is only called when k is a string, as per the type hint above.

jsonpickle/pickler.py
```diff
@@ -828,13 +828,24 @@ def _flatten_string_key_value_pair(
...
+            k.startswith(tags.JSON_KEY) or k in tags.RESERVED
+        ):
+            # Escape a data key that collides with the json:// escape prefix or
```
Theelx: Can you make this comment more concise? 8 lines is very long for a single comment.

Theelx (changes requested):
Please fix these small nitpicks and add a CHANGES.rst entry if you can. Then, I can merge it!

tests/jsonpickle_test.py
```diff
@@ -936,6 +936,28 @@ def test_string_key_requiring_escape_dict_keys_with_keys_enabled():
...
+    unpickled = jsonpickle.decode(pickled, keys=True)
+    assert unpickled == data
+
+
+def test_reserved_tag_dict_key_nested_in_object_with_keys_enabled():
+    """A reserved-tag data key survives even inside a custom object's dict field."""
+    thing = Thing("holder")
+    thing.child = {tags.OBJECT: "x", "k": 1}
+    restored = jsonpickle.decode(jsonpickle.encode(thing, keys=True), keys=True)
+    assert restored.child == {tags.OBJECT: "x", "k": 1}
```
davvid: You missed a chance to sneak a nerd joke into the test ~ replace `1` with `"cd"` for the wah wah punchline. cats are my personal go-to for these as well.

davvid (changes requested):

tests/jsonpickle_test.py
jvdillon: `tags.RESERVED` is a 20-element set, not a single item. `sorted()` gives deterministic parametrize IDs across runs (set iteration order is unstable).

tests/jsonpickle_test.py
jvdillon: Done: trimmed to one line; added a CHANGES.rst entry linking #611.

jsonpickle/pickler.py
jvdillon: Done: moved `self.keys` to the end of the condition and dropped the redundant `isinstance(k, str)`.

jsonpickle/pickler.py
jvdillon: Done: condensed to two lines.

tests/jsonpickle_test.py
jvdillon: Ha — done, the value is now "cd". 🐱

tests/jsonpickle_test.py
Theelx: Oops sorry, I forgot about that. Thanks for catching that!

Theelx (approved):

Theelx:
Oh that's very interesting with the tests. It seems that one of the typehints is incorrect, so my comment introduced a bug. Apparently k can be something other than a string, so the isinstance(k, str) guard is needed. Can you add that check back? I'll update the type hint in a separate PR. Also, for CHANGES.rst, can you change the link in parenthesis so we have ``(+611)``? ``(#611)`` will try to reference issue 611, whereas this is PR 611.

jvdillon:
Both done in 59a4f1d:

- Restored the `isinstance(k, str)` guard (verified a non-str tuple key alongside a reserved key still round-trips).
- Changed the CHANGES.rst reference to `(+611)`.

Thanks!

davvid (approved):

PR #612, opened 2026-07-09
Fix numpy view py/id desync when base is shared (#449)

### LINKED ISSUE #449: Error with strides deserializing complex NumPy array
I have found a case in which a deserialization causes the following error, using the current version in the repo (3.0.1.dev40+gef95ebc):
```
ValueError: strides is incompatible with shape of requested array and size of buffer
```

The following code shows a simple example in which that happens:
```python
from sklearn.model_selection import GridSearchCV, RepeatedStratifiedKFold
from sklearn.neighbors import KNeighborsClassifier

import numpy as np
import jsonpickle
import jsonpickle.ext.numpy

jsonpickle.ext.numpy.register_handlers()

classifier = KNeighborsClassifier()
grid = GridSearchCV(
    classifier,
    {"n_neighbors": [1, 3]},
    cv=RepeatedStratifiedKFold(),
)
grid.fit(np.eye(10), np.zeros(10))

json = jsonpickle.dumps(grid.cv_results_)
data = jsonpickle.loads(json)
```

apoorvdarshan:
---

## Summary

Fixes #449. Round-tripping a non-contiguous NumPy view whose base is shared with another view raised:

```
ValueError: strides is incompatible with shape of requested array and size of buffer
```

The maintainer confirmed an off-by-one in the encode step (a `py/id` emitted as N+1 instead of N), so the base-array reference resolved to the wrong object during restore — a 0-dim array wrapping an `unpickler._Proxy`, as reported in the issue thread.

## Root cause

When a view is stored **by reference** to a base array, `NumpyNDArrayHandlerView.flatten` also emits a decorative `values` entry, documented as *"not used in restore since base is present, but include values for human-readability"*. That decoration went through `self.context.flatten(obj.tolist(), reset=False)`, which **registers a `py/id` reference** in the pickler's bookkeeping.

Restore, however, ignores `values` entirely when a `base` is present (it rebuilds the array from `buffer`/`offset`/`strides`) and therefore never registers a matching reference. The encode and decode id counters then drift apart by one for every referenced array carrying a decorative `values`. A later view stored as `{"py/id": N}` to a shared base then resolves to the wrong object during restore, producing the strides error.

This is easy to trigger with e.g. `sklearn`'s `cv_results_` (the issue's repro): a masked array registers an unmirrored reference and the `splitN_test_score` views share a single base.

## Fix

Emit the decorative `values` **without** allocating a `py/id` slot, by snapshotting and restoring the pickler's reference bookkeeping (`_objs` / `_flattened`) around the call. This keeps the encode and decode id counters in lockstep while preserving the human-readable `values` in the output. The change is confined to the numpy extension (`jsonpickle/ext/numpy.py`); no core code is touched.

## Tests / Verification

- Added `tests/numpy_test.py::test_shared_base_after_referenced_object`, a minimal regression that reproduces the exact `ValueError` on the pristine code and passes with the fix (masked array + two non-contiguous views sharing a base).
- The original `sklearn` `cv_results_` repro from the issue now round-trips correctly with all arrays equal.
- `pytest tests/numpy_test.py tests/sklearn_test.py tests/jsonpickle_test.py` — all green (221 passed).
- Verified object-dtype views, 2D non-contiguous views, and `unpicklable=False` still round-trip correctly.
- `ruff check`, `ruff format --check`, and `isort --profile=black` are clean; `mypy --strict` introduces no new errors in the changed file.

Assisted-by: Claude (Anthropic)

Disclosure: prepared with AI assistance; reviewed and verified locally.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -37,6 +37,11 @@ v5.0.0
       ``unpicklable=False``. They now fall back to a lossy representation. (#444)
     * Fix bug where dict keys equal to a reserved wire tag (e.g. ``py/object``,
       ``py/id``) were silently dropped when encoding with ``keys=True``. (+611)
+    * Fix bug where round-tripping a non-contiguous numpy view whose base is
+      shared with another view raised ``ValueError: strides is incompatible ...``.
+      The decorative ``values`` emitted for a referenced array no longer consume a
+      ``py/id`` slot, keeping the encode and decode reference counters in sync.
+      (#449) (+612)
 
 v4.1.2
 ======
--- a/jsonpickle/ext/numpy.py
+++ b/jsonpickle/ext/numpy.py
@@ -288,6 +288,28 @@ class NumpyNDArrayHandlerView(NumpyNDArrayHandlerBinary):
         super().__init__(size_threshold, compression)
         self.mode = mode
 
+    def _flatten_values_for_readability(
+        self, obj: npt.NDArray[Any], data: dict[str, Any]
+    ) -> None:
+        """Add a human-readable values entry without touching the
+        reference bookkeeping.
+
+        When an array is stored by reference to a base, the values entry is
+        purely decorative: restore reconstructs the array from base and
+        never reads it back. Flattening the values through self.context
+        would register additional py/id references that the restore step
+        does not mirror, desynchronizing the id counters and making later
+        py/id references resolve to the wrong object (see #449). Snapshot
+        and restore the reference state so the decorative flatten is a no-op as
+        far as py/id allocation is concerned.
+        """
+        context = self.context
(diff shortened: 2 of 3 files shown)


Theelx:
Thank you for this! Sorry for not getting to it earlier. For the AI footer in the template, it was meant to be added to the commit itself. After you amend the commit to include that, I'd be happy to merge!

Theelx (approved):
Looks good once the comment about commit footers is addressed!

Theelx:
Oh also, I forgot this initially, but could you remove the backticks from the values inside the docstrings? We don't use markdown when rendering the docstrings so they're an unnecessary remnant of the AI that just clogs up the character count.

PR #614, opened 2026-07-22
Support 6-element py/reduce payloads (protocol 5 state_setter)

### LINKED ISSUE #608: decode() raises ValueError on a 6-element py/reduce payload

agu2347:
Fixes #608.

`_restore_reduce()` padded reduce tuples shorter than 5 elements, then unpacked exactly 5 names: `f, args, state, listitems, dictitems`. A 6-element reduce -- valid since pickle protocol 5 added a `state_setter` as the 6th element (PEP 307 / bpo-35900) -- can't be absorbed by that 5-name unpack and raised `ValueError: too many values to unpack (expected 5)` instead of being handled, exactly as described in the issue.

**Fix**: pad to 6 elements instead of 5, unpack a 6th `state_setter` name, and when present, call it as `state_setter(obj, state)` in place of the existing `__setstate__`/`__dict__` fallback chain, matching protocol 5's semantics: `state_setter` takes precedence over the default state-application methods when supplied.

**Testing**: verified against the issue's exact reproduction (`jsonpickle.decode('{"py/reduce": [null, null, null, null, null, null]}')`) -- no longer raises, returns `[]` same as an all-`None` 5-(or fewer)-element reduce already did. Also verified with a real 6-element reduce payload referencing actual importable functions (constructed directly against the unpickler, since jsonpickle's pickler has a separate, pre-existing 5-element limit on the *encode* side that's out of scope for this decode-side issue): the `state_setter` callable is genuinely invoked and its effect is visible on the restored object, not just tolerated. Confirmed normal 5-element (or fewer) reduce payloads without a `state_setter` are completely unaffected, still going through the existing `__setstate__`/`__dict__` fallback chain exactly as before.

Added two regression tests: one for the exact all-`None` 6-element reproduction from the issue, and one constructing a full 6-element `py/reduce` payload with dotted-path function references, asserting the `state_setter`'s effect is present after restore. I confirmed both fail with the original code (`ValueError: too many values to unpack`) and pass with the fix.

Ran the full existing test suite: 374 passed (372 baseline + 2 new), 3 skipped. The one remaining failure (`test_load_backend`, missing the optional `simplejson` package) is present identically on a clean checkout of `main`, unrelated to this change.

### DIFF
--- a/jsonpickle/unpickler.py
+++ b/jsonpickle/unpickler.py
@@ -521,9 +521,9 @@ class Unpickler:
             proxy.reset(result)
             self._swapref(proxy, result)
             return result
-        if len(reduce_val) < 5:
-            reduce_val.extend([None] * (5 - len(reduce_val)))
-        f, args, state, listitems, dictitems = reduce_val
+        if len(reduce_val) < 6:
+            reduce_val.extend([None] * (6 - len(reduce_val)))
+        f, args, state, listitems, dictitems, state_setter = reduce_val
 
         if f == tags.NEWOBJ or getattr(f, "__name__", "") == "__newobj__":
             # mandated special case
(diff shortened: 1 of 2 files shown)


Theelx:
Hello! The checklist that you deleted says that any PR that deletes the checklist will be closed immediately, as it indicates that you did not read the guidelines. Additionally, this PR was sent to fix an issue that #610 is already pending to fix, so this would be a duplicate.

PR #615, opened 2026-08-01
Fix wrong encoding for decorated function objects (#467)

### LINKED ISSUE #467: wrong encoding for decorator objects
Hi,

jsonpickle is not able to correctly decode decorated functions that return an object. 

Here's the minimal reproducible example:
```python
import jsonpickle

class FuncNested:
    def __init__(self, func) -> None:
        self.func = func

    def __call__(self, *args, **kwargs):
        return self.func(*args, **kwargs)

def nester(func):
    return FuncNested(func)


@nester
def bro():
    return "Mate"

# Check that original object is a FuncNested object
assert type(bro) is FuncNested

# Check that the `.func` member variable of the original object is NOT a FuncNested object. 
assert type(bro.func) is not FuncNested

encoded_bro_str = jsonpickle.encode(bro)

print(encoded_bro_str)
>> {"py/object": "__main__.FuncNested", "func": {"py/function": "__main__.bro"}}
decoded_bro = jsonpickle.decode(encoded_bro_str)

# Check that decoded object is also a FuncNested object
assert type(decoded_bro) is FuncNested

# Check that the `.func` member variable of decoded object is NOT a FuncNested object. 
assert type(decoded_bro.func) is not FuncNested # <-- This fails
>> ---------------------------------------------------------------------------
>> AssertionError                            Traceback (most recent call last)
>> Cell In[29], line 26
>>      24 assert type(decoded_bro) is FuncNested
>>      25 # Check that the `.func` member variable is NOT a FuncNested object. 
>> ---> 26 assert type(decoded_bro.func) is not FuncNested # <-- This fails
>> 
>> AssertionError: 
```

deepakganesh78:
Fixes #467

## Problem
When a function is decorated by a class-based decorator that replaces the module-level name with a wrapper object, `jsonpickle.encode()` produces an incorrect `py/function` reference. For example, a decorated function `bro` gets encoded as `{"py/function": "__main__.bro"}`, but `__main__.bro` now refers to the wrapper, not the original function. On decode, `decoded.func` becomes another wrapper instance instead of the original function.

## Root Cause
`util.importable_name()` uses the function's `__qualname__` to build the importable path without verifying the name still resolves to the same object. When a decorator rebinds the module-level name, the computed path points to the wrapper.

## Fix
After computing the importable name for a function, `importable_name()` now verifies the name resolves back to the original function object via the module. When it doesn't (name is shadowed), it searches common wrapper attributes (`__wrapped__`, `func`, `__func__`) on the resolved object to find the original function and returns the extended dotted path (e.g. `__main__.bro.func`) that `loadclass` can resolve correctly.

## Validation
- Regression test added in `tests/jsonpickle_test.py`
- Full test suite: **381 passed, 1 skipped** (the skip is a pre-existing optional backend test)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,7 @@
 v5.0.0
 ======
+    * Fixed encoding of decorated functions whose module-level name is rebound
+      to a wrapper object, causing incorrect decoding. (#467)
     * **Breaking Change**: The ``yaml`` module is no longer registered by default.
       You can re-enable yaml support using ``jsonpickle.ext.yaml.register()``.
       (#550) (+551)
--- a/jsonpickle/util.py
+++ b/jsonpickle/util.py
@@ -497,7 +497,23 @@ def importable_name(cls: type | Callable[..., Any]) -> str:
                 module = cls.__self__.__module__
             else:
                 module = cls.__self__.__class__.__module__
-    return f"{module}.{name}"
+    result = f"{module}.{name}"
+
+    # Verify the name resolves back to the same object.  When a function has
+    # been replaced at module level by a decorator wrapper the simple qualname
+    # points to the wrapper, not the original function.  In that case, search
+    # the wrapper for an attribute that IS the original function and return the
+    # extended dotted path so that loadclass can reach the real function.
+    if isinstance(cls, (types.FunctionType, types.BuiltinFunctionType)):
+        mod = sys.modules.get(cls.__module__)
+        if mod is not None and "." not in name:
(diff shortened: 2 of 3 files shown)


Theelx:
Hey deepak, thank you for sending this in. There was a PR template for this repo that has a line in it that says "**I understand that if this checklist is deleted from the PR body, my PR will be closed.**" This is meant to ensure that contributors actually read the requirements for a PR in order to ensure a healthy community of contributors. You deleted that checklist, so I'm going to need to close this PR.

PR #618, opened 2026-08-05
Preserve datetime fold through encode/decode

binggao1230:
### Problem

`fold` does not survive a roundtrip. `encode()` produces the *same bytes* for a
`datetime` with `fold=0` and `fold=1`:

```python
>>> dt = datetime.datetime(2023, 11, 5, 1, 30)
>>> jsonpickle.encode(dt) == jsonpickle.encode(dt.replace(fold=1))
True
```

For a tz-aware datetime in a DST fall-back hour that is not a cosmetic attribute
loss — `fold` is the only thing distinguishing the two real instants that share a
local clock reading, so decoding silently returns a different moment in time:

```python
>>> dt = datetime.datetime(2023, 11, 5, 1, 30,
...                        tzinfo=ZoneInfo("America/New_York"), fold=1)
>>> dt.astimezone(datetime.timezone.utc)
datetime.datetime(2023, 11, 5, 6, 30, tzinfo=datetime.timezone.utc)
>>> jsonpickle.decode(jsonpickle.encode(dt)).astimezone(datetime.timezone.utc)
datetime.datetime(2023, 11, 5, 5, 30, tzinfo=datetime.timezone.utc)   # an hour off
```

Nothing raises, and `restored == original` is still `True`, because `datetime.__eq__`
ignores `fold` for two operands in the same zone. Only `.utcoffset()` / `.tzname()` /
`.astimezone()` reveal it. `pickle.dumps`/`loads` roundtrips the same object correctly.

### Cause

`DatetimeHandler.flatten` used `obj.__reduce__()`, which CPython hard-codes to
`__reduce_ex__(2)`. The fold bit is encoded in the basestate (high bit of the month
byte for `datetime`, of the hour byte for `time`) only from **protocol 4** onwards —
protocols 0-3 always emit fold-less bytes for compatibility with pre-PEP-495
unpicklers:

```python
>>> dt.__reduce_ex__(2)[1][0].hex()
'07e70b05011e00000000'
>>> dt.__reduce_ex__(4)[1][0].hex()
'07e78b05011e00000000'    # byte 1: 0x0b -> 0x8b
```

`restore()` is not at fault; it faithfully rebuilds whatever bit pattern it is handed.

`Pickler._reduce` has the same problem (`obj.__reduce_ex__(2)`, commented
"we're implementing protocol 2"). It is the path taken by `datetime`/`time`
*subclasses*, which have no handler registered and are encoded as `py/reduce` —
they lose fold identically.

### Fix

Reduce at protocol 4 from both sites, via one `util.PICKLE_PROTOCOL` constant.

Protocol 4 rather than 5 deliberately: it is the lowest protocol that carries fold,
and it cannot produce PEP 574 out-of-band `PickleBuffer` objects. I diffed
`__reduce_ex__(2)` against protocols 3/4/5 over ~26 stdlib types (datetime family,
timedelta, timezone, ZoneInfo, Decimal, Fraction, UUID, set/frozenset, the
`collections` types, namedtuple, slice, range, `os.stat_result`, exceptions, plain
objects with and without `__getnewargs__`). Only three types reduce differently at
all: `datetime`, `time` — and `bytearray`/`array.array`, which never reach a reduce
path because they are intercepted earlier by `py/bytea` and `ArrayHandler`
respectively. So the encoded form changes for exactly the objects that were losing
data.

### Compatibility

Both directions verified, since `restore` never inspected the protocol:

* payloads written by older jsonpickle decode unchanged (fold-less bytes decode to
  `fold=0`, which is what they always meant) — pinned as a regression test with
  literal pre-change payloads;
* payloads written after this change decode correctly, *including fold*, on
  released jsonpickle 4.1.1.

`date` has no `fold` and its basestate is protocol-independent, so it is untouched.

### Tests

Five tests in `tests/datetime_test.py`: fold preservation for naive/aware
`datetime` and `time`; distinct encodings for the two fold values; the DST-ambiguous
hour in `America/New_York` and `Europe/Paris` asserted on `utcoffset`/`tzname`/
absolute UTC rather than `==`; the subclass path through `py/reduce`; and the legacy
payload decode. Red before, green after — reverting either call site alone leaves a
failing test. Full suite passes (453 passed, 1 skipped), and I checked the reduce
matrix and these tests on 3.10 through 3.14.

---

*Disclosure: I investigated and prepared this patch with the help of an AI assistant,
working under my direction. I read the CPython datetime pickling code, ran the
protocol diff and the compatibility checks, and verified the tests red then green
myself.*

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -49,6 +49,10 @@ v5.0.0
       (+616)
     * Fix ``bytearray`` losing its contents through encode/decode; it now
       roundtrips like ``bytes``. (+606)
+    * Fix bug where ``datetime`` and ``time`` objects lost their PEP 495 ``fold``
+      flag when roundtripped, which changed the instant a tz-aware datetime in a
+      DST fall-back hour represented. Objects are now reduced at pickle protocol
+      4 instead of 2. (+618)
 
 v4.1.2
 ======
--- a/jsonpickle/handlers.py
+++ b/jsonpickle/handlers.py
@@ -201,9 +201,9 @@ ArrayHandler.handles(array.array)
 class DatetimeHandler(BaseHandler):
     """Custom handler for datetime objects
 
-    Datetime objects use __reduce__, and they generate binary strings encoding
-    the payload. This handler encodes that payload to reconstruct the
-    object.
+    Datetime objects use __reduce_ex__, and they generate binary strings
+    encoding the payload. This handler encodes that payload to reconstruct
+    the object.
 
     """
 
@@ -215,7 +215,7 @@ class DatetimeHandler(BaseHandler):
             else:
                 result = str(obj)
             return result
-        cls, args = obj.__reduce__()  # type: ignore[str-unpack]
+        cls, args = obj.__reduce_ex__(util.PICKLE_PROTOCOL)  # type: ignore[str-unpack]
         flatten = pickler.flatten
         payload = util.b64encode(args[0])  # type: ignore[arg-type]
         args = [payload] + [flatten(i, reset=False) for i in args[1:]]
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -593,8 +593,7 @@ class Pickler:
             if has_reduce and not has_reduce_ex:
                 return obj.__reduce__()
             if has_reduce_ex:
-                # we're implementing protocol 2
-                return obj.__reduce_ex__(2)
+                return obj.__reduce_ex__(util.PICKLE_PROTOCOL)
         except TypeError:
             pass
         return None
--- a/jsonpickle/util.py
+++ b/jsonpickle/util.py
@@ -31,6 +31,9 @@ V = TypeVar("V")
 T = TypeVar("T")
 
 _ITERATOR_TYPE: type = type(iter(""))
+# Protocol used when asking objects to reduce themselves. 4 is the lowest
+# protocol that encodes datetime's PEP 495 fold bit; 2 silently drops it.
+PICKLE_PROTOCOL: int = 4
 SEQUENCES: tuple[type] = (list, set, tuple)  # type: ignore[assignment]
 SEQUENCES_SET: set[type] = {list, set, tuple}
 PRIMITIVES: set[type] = {str, bool, int, float, type(None)}
--- a/tests/datetime_test.py
+++ b/tests/datetime_test.py
@@ -36,6 +36,29 @@ class UTC(datetime.tzinfo):
 utc = UTC()
 
 
+# Payloads produced by jsonpickle before the fold flag was encoded.
+LEGACY_DATETIME = (
+    '{"py/object": "datetime.datetime", "__reduce__": '
+    '[{"py/type": "datetime.datetime"}, ["B+cLBQEeAAAAAA=="]]}'
+)
+LEGACY_DATE = (
+    '{"py/object": "datetime.date", "__reduce__": '
+    '[{"py/type": "datetime.date"}, ["B+cLBQ=="]]}'
+)
+LEGACY_TIME = (
+    '{"py/object": "datetime.time", "__reduce__": '
+    '[{"py/type": "datetime.time"}, ["AR4PAeJA"]]}'
+)
+
+
+class MyDatetime(datetime.datetime):
+    """datetime subclass; no handler is registered for it"""
+
+
+class MyTime(datetime.time):
+    """time subclass; no handler is registered for it"""
+
+
 class TimestampedVariable:
     def __init__(self, value=None):
         self._value = value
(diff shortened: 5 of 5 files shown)


Theelx:
While I understand this bug that we get by implementing pickle protocol 2 instead of 4, I'm a bit worried that changing the reduce handling to protocol 4 could break other stuff that we might not have a testcase for. I see you mentioned that the only stdlib objects that change under 4 are datetime, time, and bytearray objects, which are fine, but could there be compatibility concerns with 3rd-party code that implements 2? Alternatively, could we get away with implementing protocol 5, which has more changes but might be more future-proof? I'm just spitballing for protocol 5, but I am a bit concerned about 3rd-party code implementing pickle protocol 2 that breaks under 4.

Theelx:
While I understand this bug that we get by implementing pickle protocol 2 instead of 4, I'm a bit worried that changing the reduce handling to protocol 4 could break other stuff that we might not have a testcase for. I see you mentioned that the only stdlib objects that change under 4 are datetime, time, and bytearray objects, which are fine, but could there be compatibility concerns with 3rd-party code that implements 2? Alternatively, could we get away with implementing protocol 5, which has more changes but might be more future-proof? I'm just spitballing for protocol 5, but I am a bit concerned about 3rd-party code implementing pickle protocol 2 that breaks under 4.

binggao1230:
Fair concern, and worth being precise about what the audit does and doesn't cover.

I diffed `__reduce_ex__(2)` against 3/4/5 for ~26 stdlib types (datetime family, timedelta, timezone, ZoneInfo, Decimal, Fraction, UUID, complex, set/frozenset, bytes/bytearray, array, the collections family, slice, range, os.stat_result, plain objects). Only `datetime`/`time` change (the fold bit, from p4) and `bytearray`/`array.array` (from p3) — the latter two never actually reach this code path, they're intercepted by their own handlers first. So among what jsonpickle can encounter, protocol 4 changes exactly the two types that were losing data and nothing else.

That's stdlib-only, though. A third-party class with a hand-written `__reduce_ex__` that branches on the protocol number explicitly (`if protocol >= 4: ...`) could behave differently — I can't audit every such class, and that risk exists at any protocol bump, not just this one. What I can say: 5 buys nothing extra here (nothing reachable reduces differently between 4 and 5 — checked), so there's no reason to take on PEP 574's buffer-protocol surface for zero benefit. And `restore()` never inspects the protocol at all, so old payloads decode unchanged either way — this only affects what gets encoded going forward.

Theelx:
I'm more concerned about whether a third-party class with a hand-written reduce/reduxe_ex that *doesn't* explicitly branch on the protocol number but assumes protocol 2 could break. Is that a possibility?

Theelx:
Also, I wanted to note that in the future, any PRs that you create that don't include the checklist that you're required to have will be closed, as it indicates that you haven't read the guidelines. Because your previous PRs were contributed before that rule was in place I've made an exception so far, however I won't be doing that for future PRs.

binggao1230:
Checked: if it truly ignores the protocol arg, its output can't change — I wrote a class whose `__reduce_ex__` doesn't look at `protocol` at all, and `__reduce_ex__(2) == __reduce_ex__(4)` for it. The argument is inert to code that never reads it.

Where it could matter is the opposite case — a class that doesn't override `__reduce_ex__`/`__reduce__` at all and relies on the default machinery via `__getstate__`/`__getnewargs_ex__`, since *that* path is protocol-aware. I checked that too and saw no difference in the returned reduce tuple on this interpreter; any protocol-4-vs-2 difference there shows up in the pickler's opcode encoding, not in what gets handed to `restore()`, which (as above) never inspects protocol.

Noted on the checklist going forward, thanks.

Theelx:
Can you share the code for the class you wrote whose ``__reduce_ex__`` doesn't look at ``protocol``?

binggao1230:
Yes — it was just a scratch class like this:

```python
class IgnoresProtocol:
    def __init__(self, value=1):
        self.value = value

    def __reduce_ex__(self, protocol):
        return (type(self), (self.value,), {"extra": "state"})

obj = IgnoresProtocol(7)
assert obj.__reduce_ex__(2) == obj.__reduce_ex__(4)
```

That is the narrow point I meant: if a class implements `__reduce_ex__(protocol)` but never branches on, stores, or passes through `protocol`, asking for protocol 4 instead of 2 cannot change that class's reduce tuple.

The compatibility risk is only for classes that actually branch on `protocol`, or for stdlib types whose own reduce implementation changes by protocol. That is why I checked the stdlib reduce matrix separately and found the reachable protocol-2→4 difference only in the `datetime`/`time` fold basestate class this PR is fixing; `bytearray`/`array.array` differ too, but jsonpickle handles them before `_reduce`.

Theelx:
Sounds good, thanks for clarifying! Please resolve the merge conflict and then I'll be happy to merge.

Theelx:
I just resolved the merge conflicts myself, once the checks pass I'll merge!

PR #619, opened 2026-08-05
Rebuild numpy scalars from their declared dtype

binggao1230:
### Problem

`datetime64` and `timedelta64` scalars do not keep their unit through a roundtrip,
and for most units they do not decode at all:

```python
>>> jsonpickle.decode(jsonpickle.encode(np.datetime64("2020-06-15T12", "h"))).dtype
dtype('<M8[us]')                      # declared [h]
>>> jsonpickle.decode(jsonpickle.encode(np.timedelta64(7, "D")))
np.timedelta64(604800000000,'us')     # declared [D]
>>> jsonpickle.decode(jsonpickle.encode(np.datetime64("30000-01-01")))
ValueError: Converting an integer to a NumPy datetime requires a specified unit
```

The silent case is easy to miss because numpy compares `datetime64`/`timedelta64`
across units, so `decoded == original` stays `True` — which is all
`test_np_datetime64_units` asserted.

### Cause

`NumpyGenericHandler.restore` rebuilds the scalar with `dtype.type(value)`. The
scalar constructor takes only the kind, so numpy re-infers the unit from whatever
Python type `.tolist()` happened to produce: `datetime.date` → `D`,
`datetime.datetime` → `us`, bare `int` → the ValueError above. Everything the
dtype declares beyond its kind is dropped, even though it is sitting right there
in `data["dtype"]`.

#556 fixed the `ns` case by passing `"ns"` explicitly, on the reading that "other
units do not require special handling". Structured scalars fail on the same line
for the same reason from the other side: `np.void((1, 2.5))` wants bytes, so a
record scalar never decodes at all.

I put every numpy scalar type and every time unit through encode/decode and
compared against `pickle`, which is exact for numpy scalars: **52 of 77 cases
diverge**. Every unit of both temporal types is affected except `datetime64[ns]`,
the one #556 covered — `datetime64[D|us]` and `timedelta64[us]` survive an
in-range value by coincidence of which Python type `.tolist()` returns, but still
lose the unit on `NaT` or fail on a date outside `datetime.datetime`'s range.

### Fix

Build the scalar through the dtype instead of through `dtype.type`, so the unit
and the field layout come along, and drop the `ns` special case. The 77-case
comparison against `pickle` is then clean.

The audit turned up one more gap, on the flatten side: `longdouble` and
`clongdouble` have no Python counterpart, so `.tolist()` returns the scalar itself
and `jsonpickle.encode(np.longdouble(1.1))` recurses until the interpreter gives
up. Those two store their text form instead. Every other scalar type is
unaffected — I checked all 24 of them, `.tolist()` is Python-native for the rest.

### Tests

`test_np_datetime64_units` now loops over all 13 units and asserts the dtype
rather than just `==`; new tests cover `timedelta64` (which had no coverage at
all), values outside `datetime.datetime`'s range, structured scalars and extended
precision. All five fail on `main`. Full suite 407 passed / 5 skipped;
`ruff format` and `mypy ./jsonpickle` clean (same 5 pre-existing pandas-stubs
errors as on `main`).

---

*Disclosure: I investigated and prepared this patch with the help of an AI
assistant, working under my direction. The change is confined to the numpy
extension; I reviewed the handler, ran the differential audit against `pickle`
and verified the tests red then green myself.*

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -49,6 +49,10 @@ v5.0.0
       (+616)
     * Fix ``bytearray`` losing its contents through encode/decode; it now
       roundtrips like ``bytes``. (+606)
+    * Numpy scalars are now rebuilt from their declared dtype, so ``datetime64``
+      and ``timedelta64`` keep their unit instead of silently changing resolution
+      or failing to decode, and structured and extended-precision scalars
+      roundtrip. Generalizes the nanosecond fix from (#555). (+619)
 
 v4.1.2
 ======
--- a/jsonpickle/ext/numpy.py
+++ b/jsonpickle/ext/numpy.py
@@ -62,23 +62,25 @@ class NumpyDTypeHandler(NumpyBaseHandler):
 class NumpyGenericHandler(NumpyBaseHandler):
     def flatten(self, obj: npt.NDArray[Any], data: dict[str, Any]) -> dict[str, Any]:
         self.flatten_dtype(obj.dtype.newbyteorder("N"), data)
-        data["value"] = self.context.flatten(obj.tolist(), reset=False)
+        value = obj.tolist()
+        if isinstance(value, np.generic):
+            # extended-precision scalars have no Python counterpart, so
+            # tolist() hands back the scalar itself; store its text form
+            value = str(obj)
+        data["value"] = self.context.flatten(value, reset=False)
         return data
 
-    def restore(self, data: dict[str, Any]) -> dict[str, Any]:
+    def restore(self, data: dict[str, Any]) -> Any:
         value = self.context.restore(data["value"], reset=False)
-        return self.restore_dtype(data).type(value)  # type: ignore[no-any-return]
+        # rebuild through the dtype, not dtype.type(value): the scalar
+        # constructor ignores everything the dtype carries beyond its kind,
+        # such as the datetime64/timedelta64 unit and the field layout of a
+        # structured scalar
+        return np.array(value, dtype=self.restore_dtype(data))[()]
 
 
 class NumpyDatetimeHandler(NumpyGenericHandler):
-    """Extend NumpyGenericHandler to handle nanosecond-resolution datetime64"""
-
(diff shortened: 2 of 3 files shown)


Theelx:
This looks great, thank you! However, two of the test cases error with an "OverflowError: Overflow when converting between datetime64 units". Can you edit the times used in them such that they don't cause errors for any unit?

binggao1230:
Fixed — those two tests were breaking before jsonpickle even ran: the `astype()` calls themselves overflow int64.

`test_np_datetime64_units` used a year-2000 stamp, but ps/fs/as only span ~9s either side of the epoch — swapped it for a near-epoch stamp so it fits in every unit. `test_np_datetime64_outside_datetime_range` tried to express year 30000 in `ns`, but `ns` only spans ~584 years total (entirely inside `datetime`'s own range), so no out-of-range date can ever be expressed in it — dropped `ns` from that unit list.

Pushed (amended, same commit). Full suite green locally (407 passed, 5 skipped); CI should confirm.

Theelx:
Sounds good, I'll merge it now. However, you may want to note that there should be ~450-455 tests that run and pass with a green rather than 407 with 5 skips. I think there may be an issue with your local testing setup.

binggao1230:
You're right, thanks for catching this. My env was missing test-only extras: the pytest plugins `pytest-ruff`/`pytest-checkdocs`/`pytest-enabler` (36 of the gap is per-file ruff pseudo-tests these register), plus `pymongo`, `ecdsa`, `PyYAML`, `scipy`, `scikit-learn` (a further ~9 real tests, mostly bson/sklearn). With all of `testing` installed: 452 passed, 1 skipped (only the legit `pandas>=3.0` version-gate skip) — matches your ~450-455. None of the newly-unlocked tests touch `jsonpickle/ext/numpy.py` or datetime handling (they're bson's/sklearn's own serialization paths), so this doesn't change anything about #619 or #618 — both still pass clean under the full suite. Will make sure future envs install the complete `testing` extra.

PR #620, opened 2026-08-14
Add more extensive benchmarking suite

Theelx:

In order for us to merge your PR, please confirm that all of these boxes are true when applied to this PR and check them if so. If any box is not applicable to your PR (e.g. if you made a documentation-only change), you should just check it instead of deleting it or leaving it un-checked.

   - [x] I ensured that I included a Co-authored-by or Assisted-by tag in the footer of every commit where AI is used containing the name of all generative AI models used for that commit.
---

I've had it in my TODO list for a long time now to update the benchmarking system to make it more representative. Before this, the benchmarking system has done a good job of covering common cases and data structures, but it was limited when testing more complex situations and usage of various options in jsonpickle. This PR fixes that by hooking into pytest to use its test collection and running machinery to run benchmarks across the entire testing suite (all nearly 500 tests!). That way I don't have to duplicate code from tests if I want to reuse it for benchmarks, and it will automatically get more coverage as new tests are added.

I also added a script to benchmark various versions of jsonpickle for getting good comparison plots to visualize performance differences across versions. It currently only benchmarks all the tests that exist across all the tested versions in order to get a fair comparison, so if you don't bother testing super old versions then you'll get a more representative look since there are more tests nowadays. It also uses the geometric mean for each file so that longer tests don't skew the average like they would in an arithmetic mean.

This PR doesn't actually regenerate the current benchmarking graphs as there's no point in wasting that work until we know that this code is actually good. Once this is good enough to be merged, I'll make a followup PR that actually regenerates the graphs.

I've been working on this on and off for a few months now and it's finally ready, so I'm excited to hear what you think! While I know this is a big PR, it's also an important one so I'd appreciate a bit more than a quick skim. If you need to split up reviews across a few sessions, then that's totally fine!

### DIFF
--- a/benchmarking/README.md
+++ b/benchmarking/README.md
@@ -21,3 +21,9 @@ This core tends to be one of the lesser-used ones on your system, assuming you h
 ### Actually running the code
 You can just run ``make benchmark`` in the root of the project, and it will automatically do as much as it can to ensure a reproducible setup, and easy-to-visualize results! It should output a histogram of the results in the images/ directory.
 Another thing you can do is ``cd benchmarking && ./analyze_benchmarks.sh``. This will compare the mean speed for each benchmark between the current branch that you're working on, and the main branch, so you can check improvements or regressions.
+
+### Benchmarking the whole test suite
+``make benchmark_big`` benchmarks every test in ``tests/`` instead of the smaller suite in ``jsonpickle_benchmarks.py``, one file at a time (pass ``BENCH_JOBS=N`` to run N files at once, each pinned to its own physical core).
+
+``python3 benchmarking/benchmark_versions.py 3.0.0,3.1.0`` benchmarks your working tree against released versions in their own venvs, using only the tests that every version has in common.
+It writes one box and whisker plot per version, plus ``images/benchmark-compare-<timestamp>.png``, which reduces every shared test in the suite down to one geometric mean per version so you can see the overall change at a glance.
--- a/benchmarking/jsonpickle_pytest_benchmarks.py
+++ b/benchmarking/jsonpickle_pytest_benchmarks.py
@@ -8,6 +8,7 @@ JSONPICKLE_BENCH_JOBS to run several files concurrently.
 
 from __future__ import annotations
 
+import json
 import os
 import queue
 import re
@@ -25,7 +26,29 @@ _FILE_FLAG = "--jsonpickle-bench-file"
 _TIMESTAMP_FLAG = "--jsonpickle-bench-timestamp"
 
 
+def _item_key(item):
+    file_path = str(getattr(item, "path", "")) or item.nodeid.split("::", 1)[0]
+    return f"{Path(file_path).name}::{item.name}"
+
+
+class _CollectKeysPlugin:
+    def __init__(self):
+        self.keys = []
+
+    def pytest_collection_modifyitems(self, session, config, items):
+        self.keys.extend(_item_key(item) for item in items)
+
+
 class _BenchmarkAllTestsPlugin:
+    def __init__(self, allowed=None):
+        # for benchmark_versions we sometimes need to restrict the tests that run
+        self.allowed = allowed
+
+    def pytest_collection_modifyitems(self, session, config, items):
+        if self.allowed is None:
+            return
+        items[:] = [item for item in items if _item_key(item) in self.allowed]
+
     @pytest.hookimpl(hookwrapper=True)
     def pytest_runtest_call(self, item):
         if not isinstance(item, pytest.Function):
@@ -262,8 +285,7 @@ def _split_args(argv):
     return single_file, timestamp, extra_args
 
 
-def _run_one_file(test_file, timestamp, extra_args):
-    file_label = test_file.name
+def _pytest_config_args():
     args = []
     pytest_config = os.environ.get("JSONPICKLE_PYTEST_CONFIG")
     pytest_rootdir = os.environ.get("JSONPICKLE_ROOTDIR")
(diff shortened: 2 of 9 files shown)


davvid:
Thanks! This is really cool. I'm just starting to read/review the code tonight so I'll plan to get through it by this weekend.

Theelx:
Hey David, have you had a chance to make progress yet? No worries if not!

benchmarking/jsonpickle_pytest_benchmarks.py
```diff
@@ -314,7 +329,37 @@ def _collect_keys(test_files):
...
+
+
+def _warm_up_cpu():
+    """
+    Spin before the first measurement so that the core is already at its working clock.
+
+    On my (Theelx's) laptop, a core that has been idle starts around 40% slower and
+    takes a few hundred milliseconds to ramp up, which is long enough to inflate the
+    first test files of a run.
+    """
```
davvid, line 342: Good idea. I've also seen some instances where an extended period of hot cpu usage can cause the CPU to go into a power-savings mode, but that very much depends on factors such as whether the performance governing mode is set to "performance", whether a laptop is on battery vs plugged-in, etc.

Another strategy that can be helpful would be to filter outlier events that fall outside of the p25 and p75 range. That way a hiccup caused by some background process won't skew the numbers.

benchmarking/jsonpickle_pytest_benchmarks.py
```diff
@@ -380,6 +425,165 @@ def run_one(test_file):
     return exit_code
 
 
+def geometric_mean(values):
+    """
+    Benchmark times are on the order of 1e-5 seconds, so multiplying a few
+    hundred of them together would flush to zero long before we took the root.
+    Therefore, we compute it in log-space.
+    """
```
davvid, line 433: These are very thoughtful and useful comments, thanks for the explanations.

benchmarking/jsonpickle_pytest_benchmarks.py
```diff
@@ -380,6 +425,165 @@ def run_one(test_file):
...
+    per_file = {}
+    for path in sorted(data_dir.rglob("*.json")):
+        # every save name for this run is "<timestamp>-<file label><tag suffix>"
+        if timestamp not in path.name:
+            continue
+        try:
+            with path.open("r", encoding="utf-8") as handle:
+                payload = json.load(handle)
+        except (OSError, ValueError):
+            print(f"Ignoring unreadable benchmark data at {path}", file=sys.stderr)
```
davvid: a short `warning:` prefix on the message here can help call it out a bit.

benchmarking/README.md
```diff
@@ -21,3 +21,9 @@ This core tends to be one of the lesser-used ones on your system, assuming you h
...
 You can just run ``make benchmark`` in the root of the project, and it will automatically do as much as it can to ensure a reproducible setup, and easy-to-visualize results! It should output a histogram of the results in the images/ directory.
 Another thing you can do is ``cd benchmarking && ./analyze_benchmarks.sh``. This will compare the mean speed for each benchmark between the current branch that you're working on, and the main branch, so you can check improvements or regressions.
+
+### Benchmarking the whole test suite
+``make benchmark_big`` benchmarks every test in ``tests/`` instead of the smaller suite in ``jsonpickle_benchmarks.py``, one file at a time (pass ``BENCH_JOBS=N`` to run N files at once, each pinned to its own physical core).
+
+``python3 benchmarking/benchmark_versions.py 3.0.0,3.1.0`` benchmarks your working tree against released versions in their own venvs, using only the tests that every version has in common.
+It writes one box and whisker plot per version, plus ``images/benchmark-compare-<timestamp>.png``, which reduces every shared test in the suite down to one geometric mean per version so you can see the overall change at a glance.
```
davvid, line 29: The "using only the tests that every version has in common" is really neat.

pyproject.toml
```diff
@@ -45,7 +45,9 @@ dev = [
 testing = [
     # core
     "pytest >= 6.0, != 8.1.*",
-    "pytest-benchmark",
+    # temporarily pin to adaptive rounds branch until it's merged into main
+    # credit to @akx for implementing this!
+    "pytest-benchmark @ git+https://github.com/Theelx/pytest-benchmark.git",
```
davvid: Nicely done. I'm curious whether pypi / twine will let us publish while pinned to a git URL (cargo/rust does not), but it would be nice if it did given that these are test-only dependencies.

One slight sug here might be to create a designated branch in the pytest-benchmark fork so that we can point to that branch in the URL here. To do that we'd append `@branch=<name>` to the end of the URL immediately after `pytest-benchmark.git`. Not a merge blocker, though.

If we were extra paranoid we would maybe even pin to a specific commit ID, but this is temporary so it's all good.

davvid (approved):

davvid:
> Hey David, have you had a chance to make progress yet? No worries if not!

Thanks for the ping! Done reviewing, good stuff.

benchmarking/jsonpickle_pytest_benchmarks.py
Theelx: I considered that, but I have a thing against throwing out any data that could be useful. I've cringed at too many social scientists being okay with throwing out outliers.

pyproject.toml
Theelx: This is a great point! PyPi doesn't actually allow us to publish like this, so I've isolated it into a dependency group, which does allow us to publish like this. Thanks for bringing that up, I never thought of it and it would have broken the release.

PR #623, opened 2026-08-27
Reject ambiguous py/function resolutions during decode

fishzjp:

In order for us to merge your PR, please confirm that all of these boxes are true when applied to this PR and check them if so. If any box is not applicable to your PR (e.g. if you made a documentation-only change), you should just check it instead of deleting it or leaving it un-checked.

   - [x] I ensured that I included a Co-authored-by or Assisted-by tag in the footer of every commit where AI is used containing the name of all generative AI models used for that commit.
---

Decoding a ``py/function`` tag whose name is shadowed by a non-function binding (e.g. a decorated function whose name now points at its own wrapper) silently returns that object ([#467](https://github.com/jsonpickle/jsonpickle/issues/467)).

``_restore_function`` accepted anything ``util.loadclass()`` resolved. It now only returns functions, classes, or ``None`` (the existing miss semantics); anything else raises ``TypeError``, so ambiguous references fail loudly instead of decoding into a wrong-typed object.

Tests: added a regression test for the wrapper-shadowing case in #467, with the payload taken straight from ``encode()``. Full suite: 396 passed, 2 skipped; ``ruff check`` / ``ruff format --check`` clean.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -59,6 +59,9 @@ v5.0.0
       roundtrip. Generalizes the nanosecond fix from (#555). (+619)
     * The ability to run benchmarks over the entire pytest suite has been added
       in order to cover a wider portion of the codebase with the benchmarks. (+620)
+    * Decoding a ``py/function`` tag whose name resolves to an existing
+      non-function binding now raises ``TypeError`` instead of returning
+      the object unchanged. (#467) (+623)
 
 v4.1.2
 ======
--- a/jsonpickle/unpickler.py
+++ b/jsonpickle/unpickler.py
@@ -5,6 +5,7 @@
 # This software is licensed as described in the file COPYING, which
 # you should have received as part of this distribution.
 import dataclasses
+import inspect
 import warnings
 from collections.abc import Callable, Iterator, Sequence
 from typing import Any, TypeAlias
@@ -891,7 +892,13 @@ class Unpickler:
         return self._restore_object_instance(obj, cls, class_name)
 
     def _restore_function(self, obj: dict[str, Any]) -> Any:
-        return util.loadclass(obj[tags.FUNCTION], classes=self._classes)
+        class_name = obj[tags.FUNCTION]
+        result = util.loadclass(class_name, classes=self._classes)
+        if inspect.isroutine(result) or inspect.isclass(result) or result is None:
+            return result
+        raise TypeError(
+            f"{class_name!r} resolves to a {type(result).__qualname__}, not a function"
+        )
 
     def _restore_set(self, obj: dict[str, Any]) -> set[Any]:
         try:
--- a/tests/jsonpickle_test.py
+++ b/tests/jsonpickle_test.py
@@ -42,6 +42,24 @@ class Thing:
         return 'Thing("%s")' % self.name  # ruff: ignore[UP031]
 
 
+class AmbiguousDecorator:
+    def __init__(self, func):
+        self.func = func
+
+    def __call__(self, *args, **kwargs):
+        return self.func(*args, **kwargs)
+
+
+def _ambiguous_factory(func):
+    return AmbiguousDecorator(func)
+
+
+# ambiguous_target is bound to the wrapper instance, shadowing the function
+@_ambiguous_factory
+def ambiguous_target():
+    return "Mate"
+
+
 class Capture:
     def __init__(self, *args, **kwargs):
         self.args = args
@@ -728,6 +746,20 @@ def test_builtin_function():
     actual = jsonpickle.decode(json)
     assert expect == actual
     assert expect is actual
+
+
+def test_ambiguous_function_reference_regression_issue467():
+    assert type(ambiguous_target.func).__name__ == "function"
+    encoded = jsonpickle.encode(ambiguous_target)
+    # the payload must come from encode(); hand-built wire tags get stripped
+    assert tags.FUNCTION in encoded
+    with pytest.raises(TypeError):
+        jsonpickle.decode(encoded)
+
+
+def test_function_reference_resolution_contract():
+    assert jsonpickle.decode('{"py/function": "builtins.dir"}') is dir
+    assert jsonpickle.decode('{"py/function": "no.such.module.attr"}') is None
 
 
 def test_restore_legacy_builtins():


jsonpickle/unpickler.py
```diff
@@ -891,7 +892,13 @@ def _restore_object(self, obj: dict[str, Any]) -> Any:
         return self._restore_object_instance(obj, cls, class_name)
 
     def _restore_function(self, obj: dict[str, Any]) -> Any:
-        return util.loadclass(obj[tags.FUNCTION], classes=self._classes)
+        class_name = obj[tags.FUNCTION]
+        result = util.loadclass(class_name, classes=self._classes)
+        if inspect.isroutine(result) or inspect.isclass(result) or result is None:
+            return result
```
davvid, line 898: Is this check equivalent to `if result is None or callable(result):` ?

It'd be simpler to write it that way if so. I'm in favor of the flexibility of allowing any callable to be returned here (assuming `callable` is a superset of `isroutine | isclass`) as users w/ custom handlers could theoretically be relying on jsonpickle being flexible in what it accepts here. Being strict about it at least being `callable` seems like a good constraint to me, tho..

tests/jsonpickle_test.py
```diff
@@ -730,6 +748,20 @@ def test_builtin_function():
     assert expect is actual
 
 
+def test_ambiguous_function_reference_regression_issue467():
```
davvid, line 751: nit: let's not name tests after issue numbers. Better for the code to be succinct and self-contained so that we don't need to jump anywhere else to understand what 467 means.

Naming-wise, I'd suggest `def test_non_callable_raises` and `def test_callable_reference` for the other test below.

tests/jsonpickle_test.py
```diff
@@ -730,6 +748,20 @@ def test_builtin_function():
...
+    assert type(ambiguous_target.func).__name__ == "function"
```
davvid, line 752: I don't think we need this assert. In a `callable`-flavored implementation we might `assert callable(ambiguous_target)` but that seems kinda tautological and not worth asserting IMO.

jsonpickle/unpickler.py
```diff
@@ -891,7 +892,13 @@ def _restore_object(self, obj: dict[str, Any]) -> Any:
         return self._restore_object_instance(obj, cls, class_name)
 
     def _restore_function(self, obj: dict[str, Any]) -> Any:
-        return util.loadclass(obj[tags.FUNCTION], classes=self._classes)
+        class_name = obj[tags.FUNCTION]
```
davvid, line 895: even though we use `loadclass` we don't really expect a class to be returned. We expect a callable, so renaming these variables would make sense. sug: `callable_name` instead of `class_name`.

jsonpickle/unpickler.py
```diff
@@ -891,7 +892,13 @@ def _restore_object(self, obj: dict[str, Any]) -> Any:
...
+        result = util.loadclass(class_name, classes=self._classes)
+        if inspect.isroutine(result) or inspect.isclass(result) or result is None:
+            return result
+        raise TypeError(
+            f"{class_name!r} resolves to a {type(result).__qualname__}, not a function"
+        )
```
davvid, line 901:
```python
f"{callable_name!r} is not a callable object: {result!r}"
```

Also, maybe splitting hairs, but it almost seems like ValueError would be better to raise.

In general, everything kind of a is a value error because we're rejecting the value they passed us.

We don't actually document the exceptions we return. That seems like something we might want to nail down (thoughts @Theelx?)

My gut feeling would probably want to lean on ValueError for the majority of our error scenarios. If we're inconsistent about this it seems like it'd be easy to rectify, at least.

CHANGES.rst
```diff
@@ -59,6 +59,9 @@ v5.0.0
       roundtrip. Generalizes the nanosecond fix from (#555). (+619)
     * The ability to run benchmarks over the entire pytest suite has been added
       in order to cover a wider portion of the codebase with the benchmarks. (+620)
+    * Decoding a ``py/function`` tag whose name resolves to an existing
+      non-function binding now raises ``TypeError`` instead of returning
+      the object unchanged. (#467) (+623)
```
davvid, line 64: Thanks for adding a changelog entry as well.

davvid (changes requested):
Besides the various tangents, this is on the right track. Correct the naming, exception choice, and the callable implementation details and we'll can merge IMO.

tests/jsonpickle_test.py
```diff
@@ -730,6 +748,20 @@ def test_builtin_function():
...
+    encoded = jsonpickle.encode(ambiguous_target)
+    # the payload must come from encode(); hand-built wire tags get stripped
```
davvid, line 754: this comment doesn't really mean anything. It's better to just remove it IMO.

Hmm.. I'm kinda confused now. I don't think jsonpickle should produce an encoded payload that would result in throwing an exception when handed right back to decode() like this test does.

That's something we shouldn't do full-stop. https://en.wikipedia.org/wiki/Robustness_principle

We should only throw an exception if the user hand-crafted a payload that results in a non-callable.

It seems like a test similar to the `decode(str)` ones below where we get it to resolve a non-callable would be better for testing the error scenario.

Apologies if I'm missing something about why we're trying to disallow these objects. I've got a little bit of an extra 👁️ on this now that this test has me confused about the intent of the changes.

Could you please write a commit message that that helps justify the changes, explaining *why* we need these changes? Please not an AI one, either, a few regular sentences from you would be ideal.

That would also be best on a single commit, so squashing both commits together would make sense. We don't need a separate commit for the test case -- it's better to have the fix and the tests all in one.

davvid:
Just catching up -- sorry I hadn't read the linked issues recently and had forgotten that this is the other side of an error scenario for a submitted issue.

Well, I don't really think this is the right answer. It seems like this is really an encoding bug (insofar as we don't currently have enough fidelity to write down enough breadcrumbs while encoding to reconstruct equivalent objects while decoding).

In that respect, I take it all back -- I don't think this is something we should merge. A better solution would be to actually make the roundtrip succeed.

If that can't happen, then *maybe* an opt-in flag to disallow these objects during the encode() step would at least be a mitigation that prevents the issues at the source.

Along those lines of "fix it upstream in encode()" would be a consideration of what kind of tool we might give users for dealing with these objects.

Having them degrade to `None` / `nil` in the encoded payload might be a practical approach, for example, for a user that might be using a subset of an object's fields for serialization purposes.

All that said, I'm going to close this for now since this seems a bit off-target from a solution, even though the intent is well-meaning. Please feel free to open a new PR if you'd like to tackle this from the other side of the encode() / decode() bridge.

This just seems slightly restrictive from decode()'s perspective, especially if there's the possibility of exotic use cases that we might be shutting the door on by adding these constraints.

fishzjp:
Thanks for taking the time to catch up on this one, and for the clear write-up — I agree the missing fidelity lives on the encode() side of the bridge, and that decode() shouldn't be restrictive about how it resolves references.
Thanks again for the direction!

PR #624, opened 2026-09-02
Warn instead of silently dropping dict keys that collide with reserved tags

Ashishjob:
## Problem

In the default `keys=False` mode, a plain-string dict key equal to a reserved jsonpickle wire tag (e.g. `"py/object"`, `"py/id"`, `"py/tuple"`) is dropped by the pickler:

```pycon
>>> import jsonpickle
>>> jsonpickle.encode({"py/object": "hello", "normal": 1})
'{"normal": 1}'          # the "py/object" entry is gone
```

This is intentional in the sense that such a key cannot be represented in `keys=False` output — it would be misread as that tag on decode, and `keys=False` deliberately passes `json://`-prefixed keys through verbatim (see `test_dict`), so there is no room to escape it. But the value is currently lost **silently**, which is an easy footgun for anyone whose data happens to contain such keys (config blobs, API payloads, ML metadata, etc.).

## Change

Emit a `UserWarning` pointing at `keys=True` (which escapes and preserves these keys) instead of dropping the value with no indication. This is scoped precisely:

- Only fires for a **string** key that is exactly a reserved tag, and only on the `keys=False` path where the drop actually happens.
- The drop behaviour itself is unchanged, so `keys=True` and any non-colliding key are entirely unaffected.

```pycon
>>> jsonpickle.encode({"py/object": "hello"})
UserWarning: jsonpickle: dropping dict key 'py/object' that collides with a reserved tag; use keys=True to preserve it
>>> jsonpickle.decode(jsonpickle.encode({"py/object": "hi"}, keys=True), keys=True)
{'py/object': 'hi'}
```

## Tests

Added `test_dict_reserved_key_warns_when_dropped` (warns + still drops in `keys=False`; preserved and no warning with `keys=True`). Verified it fails before the change and passes after. Full `tests/jsonpickle_test.py` passes except a pre-existing `test_load_backend` failure unrelated to this change (missing optional JSON backend in my env).

Happy to adjust — e.g. if you would prefer to raise, or escape reserved keys in `keys=False` too, I can take it in that direction instead.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,8 @@
 v5.0.0
 ======
+    * jsonpickle now warns instead of silently dropping a dict key that collides
+      with a reserved tag (e.g. ``"py/object"``) when encoding with the default
+      ``keys=False``; use ``keys=True`` to preserve such keys. (+624)
     * **Breaking Change**: The ``yaml`` module is no longer registered by default.
       You can re-enable yaml support using ``jsonpickle.ext.yaml.register()``.
       (#550) (+551)
--- a/jsonpickle/pickler.py
+++ b/jsonpickle/pickler.py
@@ -465,6 +465,18 @@ class Pickler:
         self, k: Any, v: Any, data: dict[str | Any, Any]
     ) -> dict[str | Any, Any]:
         """Flatten a key/value pair into the passed-in dictionary."""
+        if isinstance(k, str) and k in tags.RESERVED:
+            # A plain-string dict key that collides with a reserved jsonpickle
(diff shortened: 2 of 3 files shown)


Theelx:
Hey ashishjob, thank you for sending this in. There was a PR template for this repo that has a line in it that says "I understand that if this checklist is deleted from the PR body, my PR will be closed." This is meant to ensure that contributors actually read the requirements for a PR in order to ensure a healthy community of contributors. You deleted that checklist, so I'm going to need to close this PR.

PR #627, opened 2026-09-05
Fix roundtrip of compiled bytes regex patterns

dualfroz:
## Problem

Encoding and decoding a compiled `bytes` regex pattern does not roundtrip.

```python
import re, jsonpickle
p = re.compile(rb"[a-z]+", re.IGNORECASE)
d = jsonpickle.decode(jsonpickle.encode(p))
type(d.pattern)      # <class 'str'>  -- was <class 'bytes'>
d.match(b"abc")      # TypeError: cannot use a string pattern on a bytes-like object
```

With the default backend chain (`simplejson`/`ujson`) the bytes payload is
silently coerced to text, so the decoded object is a *text* pattern that no
longer matches bytes input. With the stdlib `json` backend encoding raises
`TypeError: Object of type bytes is not JSON serializable` outright.

## Root cause

`jsonpickle/handlers.py`, `RegexHandler.flatten` stored the raw pattern:

```python
data["pattern"] = obj.pattern
```

For a bytes pattern `obj.pattern` is `bytes`, which is not a JSON string. The
value is written verbatim into the flattened dict and handed to the JSON
backend, which either raises or coerces the bytes to text. `RegexHandler.restore`
then calls `re.compile(data["pattern"], ...)` on that text and produces a text
pattern, losing both the byte type and (because compiling a text pattern adds
`re.UNICODE` automatically) the original flags.

## Fix

`jsonpickle/handlers.py`: detect a bytes pattern in `flatten`, store its payload
as base64 with an `is_bytes` marker, and decode it back to bytes in `restore`.
Text patterns are unchanged, so old payloads (which have no `is_bytes` key)
continue to decode exactly as before.

## Test

`tests/object_test.py::test_compiled_bytes_regex_roundtrip` encodes and decodes
a `bytes` pattern through the full JSON pipeline and asserts the restored pattern
is still `bytes`, keeps its flags, and matches bytes input. It fails on the
current `main` (`AssertionError: assert isinstance('[a-z]+', bytes)`) and passes
with the fix. The existing `test_thing_with_compiled_regex` and
`test_compiled_regex_flags_roundtrip` still pass.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -59,6 +59,11 @@ v5.0.0
       roundtrip. Generalizes the nanosecond fix from (#555). (+619)
     * The ability to run benchmarks over the entire pytest suite has been added
       in order to cover a wider portion of the codebase with the benchmarks. (+620)
+    * Fix bug where compiled ``bytes`` regex patterns (e.g. ``re.compile(b"...")``)
+      failed to roundtrip: encoding raised with the stdlib ``json`` backend and
+      silently degraded the pattern to text with ``simplejson``/``ujson``, so the
+      decoded object was a text pattern that could no longer match bytes input.
+      The pattern payload is now preserved as base64 and rebuilt as bytes.
 
 v4.1.2
 ======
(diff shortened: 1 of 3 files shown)


Theelx:
Hey dualfroz, thank you for sending this in. There was a PR template for this repo that has a line in it that says "I understand that if this checklist is deleted from the PR body, my PR will be closed." This is meant to ensure that contributors actually read the requirements for a PR in order to ensure a healthy community of contributors. You deleted that checklist, so I'm going to need to close this PR.

PR #632, opened 2026-09-22
Fix silent data loss when encoding a memoryview

itzzdev09:
Encoding a `memoryview` silently discards its contents:

```python
>>> import jsonpickle
>>> jsonpickle.encode(memoryview(b"hello world"))
'{"py/object": "builtins.memoryview"}'
>>> jsonpickle.decode(_)
{'py/object': 'builtins.memoryview'}
```

No error, no warning — the buffer's bytes are just gone, and decoding doesn't even reconstruct a `memoryview`, only the raw stub dict, since `memoryview` can't be instantiated through the generic reflection path the same way ordinary objects can.

`bytes` and `bytearray` already have dedicated handling that survives a round trip; `memoryview` fell through to the generic object path because it has no useful `__reduce__`. This adds the same treatment as `bytearray`: a `py/memoryview` tag wrapping the base64/base85-encoded bytes, restored as `memoryview(data)`.

Scope: like `bytearray`, this always round-trips through raw bytes, so a `memoryview` over a non-byte buffer (e.g. `memoryview(array.array('i', [...]))`, itemsize > 1) restores with the original content but as a plain byte-format view rather than preserving `itemsize`/`format`. That's a real improvement over the current 100% data loss either way.

Added `test_memoryview_roundtrip`, `_empty_roundtrip`, `_nested_roundtrip`, `_base64_default`, `_base85`, `_base85_roundtrip`, mirroring the existing `bytearray` tests. 6 of 7 fail on `main`; full suite passes (417/419 — the 2 failures are pre-existing on `main`, unrelated to this change).


Theelx:
Hey itzzdev09, thank you for sending this in. There was a PR template for this repo that has a line in it that says "I understand that if this checklist is deleted from the PR body, my PR will be closed." This is meant to ensure that contributors actually read the requirements for a PR in order to ensure a healthy community of contributors. You deleted that checklist, so I'm going to need to close this PR.

